PR #22, opened 2024-08-10
 feat: add minify option

ryoppippi:
### Description
Add minify option for output

### Additional context
https://github.com/rolldown/rolldown/blob/a2bb2cae97e0f79f712c8ff68d116366ee4ecf35/packages/rolldown/src/options/output-options.ts#L48

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -25,6 +25,8 @@ export interface Options {
   alias?: Record<string, string>
   /** @default true */
   treeshake?: boolean
+  /** @default true */
+  minify?: boolean
   /** @default 'node' */
   platform?: 'node' | 'neutral'
   /**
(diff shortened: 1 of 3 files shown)


src/options.ts
```diff
@@ -65,6 +67,7 @@ export async function normalizeOptions(
     platform = 'node',
     outDir = 'dist',
     dts = false,
+    minify = true,
```
sxzz: I think it shouldn't be minified by default, just like tsup.

```suggestion
    minify = false,
```

src/options.ts
```diff
@@ -23,6 +23,8 @@ export interface Options {
   alias?: Record<string, string>
   /** @default true */
   treeshake?: boolean
+  /** @default true */
```
```suggestion
  /** @default false */
```

src/options.ts
ryoppippi: Thanks

PR #28, opened 2024-08-23
feat: add silent option

ryoppippi:
### Description

add `silent` option just like `tsup` has

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -18,6 +18,7 @@ export async function runCLI(): Promise<void> {
     })
     .option('--clean', 'Clean output directory')
     .option('--minify', 'Minify output')
+    .option('--silent', 'Suppress non-error logs')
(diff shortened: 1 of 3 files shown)


sxzz (approved):

sxzz:
Made some tweaks via https://github.com/sxzz/tsdown/commit/4c9312bd7d39a38489a50f179a107c58da88180c

ryoppippi:
@sxzz I'm developing a new library, and inside the library calls `build` function from `tsdown`, like this:
```ts
import { build } from '@sxzz/tsdown';
```

How can I silent the log in this case?

sxzz:
I exported `logger` for this case
https://github.com/sxzz/tsdown/commit/56efd13c29a678315967bd9aecc94486e6624e6b

ryoppippi:
Oh, Thanks! |'ll use that

PR #96, opened 2025-04-16
chore: add usage information in readme [ci skip]

Gugustinette:
Hi !

It can be obvious if you're already into building tools, but I missed a basic example of how to use tsdown (specially since there's still no documentation).

"Install", "Configuration" and "Usage" sections could also be merged into a single "Installation" section, using bullet points for each part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,12 @@ export default defineConfig({
 })
 ```
 
+## Usage
+
+```ts
+npx tsdown
+```
+
 ## Migrate from tsup
 
 ```bash


sxzz (approved):

Gugustinette:
@sxzz 

just realized I used \`\`\`ts
Capture d’écran 2025-04-16 à 18 09 49

instead of \`\`\`bash
Capture d’écran 2025-04-16 à 18 09 34

so it displays an error in the IDE lol my bad 😭

PR #105, opened 2025-04-17
ci: improve workflows

okineadev:
### Description

I improved the security of workflows by restricting permissions and credentials, before running `zizmor` to check for problems

I also added emojis to the workflow names because they make navigation much better and it's easier to understand what they do

Also simplified them to reduce their usage


okineadev:
> I prefer to keep `pkg-pr-new` in a separate workflow

Is there a good reason for this?

Personally, I changed it for speed

sxzz:
`pkg-pr-new` will run multiple times due to the matrix setup. Additionally, we want it to always trigger a release, regardless of whether linting and unit tests pass or fail.

okineadev:
Oh, okay, I'll fix it tomorrow

okineadev:
> `pkg-pr-new` will run multiple times due to the matrix setup. Additionally, we want it to always trigger a release, regardless of whether linting and unit tests pass or fail.

Fixed

lishaduck:
> before running `zizmor` to check for problems

You listen to me now? :P

okineadev:
> > before running `zizmor` to check for problems
> 
> You listen to me now? :P

~I was kicked out of the organization and migrated to Albania for some reason~

But seriously, this is a cool tool, better than CodeQL for actions

PR #123, opened 2025-04-20
chore: update `config-options.md`

jsparkdev:
- update the `config-options` pages.
- let me know if you have more information to add (like a description of the options).
- please point out any incorrect values (like default values).


Gugustinette:
`config-options.md` is generated using the [docs-generate.sh](https://github.com/rolldown/tsdown/blob/main/docs/.vitepress/scripts/docs-generate.sh) script, so changing its content directly should be done by editing the script.

This means you should find a reliable way to do modifications inside the markdown file from the bash script.

jsparkdev:
@Gugustinette 

Thanks for the new information!

I'll close this PR.

PR #136, opened 2025-04-21
feat: report size

### LINKED ISSUE #116: feat: total output size after build
It's great that it tells you the sum of the last files after each file in Unbuild. Because that way it's easy to see where we made a mistake or if we added something wrong to the package.

MatteoGabriele:
### Description
This PR adds a size reporting feature that calculates and displays the sizes of bundled files in the output directory. 

The implementation:
- Identifies and lists type declaration files (.d.ts, .d.mts, .d.cts)
- Calculates the raw, gzipped, and brotli size of each bundle file
- Computes and displays the total size across all bundle files
- Uses human-readable formatting for file sizes

### Linked Issues
fix #116 

### Additional context
You might want to consider a different style and format for the logs, and it could be better to make this an optional feature instead of the default.

I hope I was able to help a little. Feel free to use the code as you see fit, or not at all. 

Thank you for your work!


src/features/sizes.ts
```diff
@@ -0,0 +1,80 @@
...
+import { formatBytes } from '../utils/format'
+import { logger } from '../utils/logger'
+
+const debug = Debug('tsdown:sizes')
+
+const DTS_EXTENSIONS = ['.d.ts', '.d.mts', '.d.cts']
+
+interface SizeInfo {
+  size: number
+  gzip: number
```
TheAlexLichter: I'd advocate for supporting `brotli` too!

MatteoGabriele: @TheAlexLichter I have just pushed changes to include brotli support as well

sxzz (approved):

sxzz:
Refactored a lot, thanks!

PR #145, opened 2025-04-23
chore: add `rolldown`/`unrs-resolver` to ignoredBuiltDependencies

btea:
### Description
The preinstall and postinstall scripts for both `rolldown` and `unrs-resolver` can be disabled.


sxzz (commented):
Please revert empty line.

btea:
Done. I think there is something wrong with my local environment.

PR #146, opened 2025-04-23
fix: watch mode outDir cannot be an empty string

btea:
### Description
When `outDir` is an empty string, the project root directory is added to the ignored list, which will invalidate the watch.

### DIFF
--- a/src/features/watch.ts
+++ b/src/features/watch.ts
@@ -12,18 +12,21 @@ export async function watchBuild(
   rebuild: () => void,
   restart: () => void,
 ): Promise<FSWatcher> {
-  const root = process.cwd()
-  if (options.outDir === root) {
-    throw new Error(
-      'options.outDir cannot be an empty string, which will cause the watch to be invalid.',
-    )
-  }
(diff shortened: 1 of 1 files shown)


src/features/watch.ts
```diff
@@ -12,18 +12,21 @@ export async function watchBuild(
...
-  if (options.outDir === root) {
-    throw new Error(
-      'options.outDir cannot be an empty string, which will cause the watch to be invalid.',
-    )
-  }
   const { watch } = await import('chokidar')
   const debouncedRebuild = debounce(rebuild, 100)
 
+  const cwd = process.cwd()
+  if (typeof options.watch === 'boolean' && options.outDir === cwd) {
```
btea, line 19: This judgment is more appropriate to be placed above. If the conditions are met, an error will be thrown directly without importing related packages. 🤔

sxzz: Thanks! 516fab0f5b2414b177851a5aaf033d1dbd1687f4

btea: So efficient! 👍

PR #147, opened 2025-04-24
feat: add `--debug` option

btea:
### Description
Support `--debug` option.


sxzz:
Is there any reason to have the flag?

btea:
It is convenient for users to view the debug log, which may be helpful for debugging some problems in the future.

image

sxzz:
Since this feature is not widely or frequently used by typical users, I don't think it's necessary to introduce this flag. Additionally, it could slow down cold startup.

sxzz:
Also users can use add environment variables manually. `DEBUG=tsdown`

src/run.ts
```diff
@@ -1,4 +1,29 @@
 #!/usr/bin/env node
-import { runCLI } from './cli'
+import process from 'node:process'
 
-runCLI()
+const debugIndex: number = process.argv.findIndex((arg) =>
+  /^--debug$/.test(arg),
+)
+
+if (debugIndex > 0) {
```
sxzz: Can we read `debug` flag from cac?

src/index.ts
```diff
@@ -42,8 +45,14 @@ export async function build(userOptions: Options = {}): Promise<void> {
 
   debug('Loading config')
   const { configs, file: configFile } = await resolveOptions(userOptions)
-  if (configFile) debug('Loaded config:', configFile)
-  else debug('No config file found')
+  if (configFile) {
+    debug('Loaded config:', configFile)
+    configs.forEach((config) => {
+      debug('using resolved config: %O', { ...config })
```
sxzz: Why `{ ...config}` here?

src/index.ts
```diff
@@ -27,11 +28,13 @@ import {
   type ResolvedOptions,
 } from './options'
 import { ShebangPlugin } from './plugins'
-import { debug, logger, setSilent } from './utils/logger'
+import { logger, setSilent } from './utils/logger'
 import { prettyFormat, readPackageJson } from './utils/package'
+
```
sxzz: Revert blank line

btea:
> Additionally, it could slow down cold startup.

If the user does not configure this option, there should be no difference from the original, so there should be no problem. If the option is configured, there is a corresponding demand, so a slightly slower speed may be understandable.

btea:
> Also users can use add environment variables manually. `DEBUG=tsdown:*`

Indeed, but it seems difficult to configure environment variables uniformly in different terminals. It is more convenient to add an option.

src/run.ts
btea: The current structure does not seem to work. The packaged code will be imported into debug before cac runs, which may cause the configured environment variables to fail to take effect.

src/index.ts
btea: It's just a subconscious writing habit.

src/index.ts
btea: Ok.

src/run.ts
sxzz: We don't need to debug cac, only should ensure set DEBUG before tsdown runs

btea: Yes, so I think it is better to put it at the top of the project entrance.

sxzz (approved):

PR #152, opened 2025-04-25
feat(cli): support compile-time env

### LINKED ISSUE #138: Add support for compile-time environment variables
`tsup` supports [compile-time environment variables](https://tsup.egoist.dev/#compile-time-environment-variables), which allow you to specify values for environment variables using the `--env` flag:

```
tsup --env.NODE_ENV production
```

I think you can achieve a similar result by using `inputOptions` and `NODE_ENV=production tsdown`:

```ts
export default defineConfig(() => {
  return {
    inputOptions: {
      define: {
        'process.env.NODE_ENV': process.env.NODE_ENV ?? undefined,
      },
    }
  };
});
```

However, it would be nice to have the same CLI flag to achieve this without modifying the config.

fu050409:
### Description

Support compile-time env in command-line tool

### Linked Issues

Resolved #138

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -152,12 +152,12 @@ export interface Options {
    * @example
    * ```ts
    * {
-   *   "DEBUG": "true",
+   *   "DEBUG": true,
    *   "NODE_ENV": "production"
    * }
    * ```
    */
-  env?: Record<string, string>
+  env?: Record<string, any>
 
   hooks?:
     | Partial<TsdownHooks>
--- a/src/options.ts
+++ b/src/options.ts
@@ -150,10 +150,12 @@ export interface Options {
   /**
    * Compile-time env variables.
    * @example
+   * ```ts
    * {
    *   "DEBUG": "true",
    *   "NODE_ENV": "production"
    * }
+   * ```
    */
   env?: Record<string, string>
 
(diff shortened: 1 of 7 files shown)


src/options.ts
```diff
@@ -147,6 +147,16 @@ export interface Options {
    */
   report?: boolean | ReportOptions
 
+  /**
+   * Compile-time env variables.
+   * @example
+   * {
+   *   "DEBUG": "true",
```
```suggestion
   *   "DEBUG": true,
```

sxzz (approved):
Thanks!

src/options.ts
fu050409: @sxzz `env` is parsed by `cac` in the terminal, and if you use `--env.DEBUG=true`, the value here is actually a string `"true"` rather than a boolean. In Node.js, `process.env` is defined as extending `Dict<string>`, so I think it should look like a string here rather than a boolean. Wdyt?

sxzz: Here is for JavaScript API. How to have a boolean type `true` in your case? We should add the case into unit test

sxzz: In cac, the value will be `true` by using `--env.X` without value

fu050409: > In cac, the value will be `true` by using `--env.X` without value

Thanks for pointing out.

> Here is for JavaScript API. How to have a boolean type `true` in your case? We should add the case into unit test

I'll make the change.

sxzz (approved):
Thanks!

Issue #127, opened 2025-04-20
Built-in stubbing


lishaduck:
I know there's https://github.com/sxzz/stub-demo, but it'd be nice if that was built-in (like in unbuild).

PR #164, opened 2025-04-27
feat: add stubbing feature

Gugustinette:
### Description

Add a stubbing feature, similar to the [unbuild](https://github.com/unjs/unbuild?tab=readme-ov-file#-passive-watcher) one.

### Linked Issues

#127 

### Additional context

Hi @sxzz ! I tried something for the stubbing feature (based on [your stub demo](https://github.com/sxzz/stub-demo)), but this is my first time playing with the tsdown codebase so it is very likely I don't know what I'm doing at all lol

I'd be very happy to get some feedbacks on this !

I guess it lakes a way of doing some kind of JIT transpilation, using [jiti](https://github.com/unjs/jiti) as unbuild does :
```js
// dist/index.mjs
import jiti from 'jiti'

export default jiti(null, { interopDefault: true })('/Users/antfu/unbuild-test/src/index')
```

But first I was mostly asking for a feedback on the overall way of editing the tsdown code 😭


src/features/stubbing.ts
```diff
@@ -0,0 +1,41 @@
+import path from 'node:path'
```
sxzz: Rename to `stub.ts`

src/features/stubbing.ts
```diff
@@ -0,0 +1,41 @@
...
+import type {
+  NormalizedOutputOptions,
+  OutputBundle,
+  Plugin,
+  PluginContext,
+} from 'rolldown'
+
+export function StubbingPlugin(entry: Options['entry']): Plugin {
+  return {
+    name: 'tsdown:stubbing',
```
sxzz:
```suggestion
    name: 'tsdown:stub',
```

src/features/stubbing.ts
```diff
@@ -0,0 +1,41 @@
+import path from 'node:path'
+import type { Options } from '../options'
+import type {
+  NormalizedOutputOptions,
+  OutputBundle,
+  Plugin,
+  PluginContext,
+} from 'rolldown'
+
+export function StubbingPlugin(entry: Options['entry']): Plugin {
```
sxzz:
```suggestion
export function StubPlugin(entry: Options['entry']): Plugin {
```

src/features/stubbing.ts
```diff
@@ -0,0 +1,41 @@
...
+  return {
+    name: 'tsdown:stubbing',
+    generateBundle(
```
sxzz: I think use `load` hook will be much better

src/index.ts
```diff
@@ -244,6 +246,10 @@ async function getBuildOptions(
     plugins.push(ShebangPlugin(cwd))
   }
 
+  if (stub) {
```
sxzz, line 248: We should disable some plugins on stub mode, like dts, report, external, unused...

src/options.ts
```diff
@@ -85,6 +85,8 @@ export interface Options {
   define?: Record<string, string>
   /** @default false */
   shims?: boolean
+  /** @default false */
+  stub?: boolean
```
sxzz, line 116: Should also update `ResolvedConfig` for fixing types

tests/index.test.ts
```diff
@@ -256,3 +256,32 @@ test('minify', async (context) => {
   expect(snapshot).not.contains('true')
   expect(snapshot).not.contains('const foo')
 })
+
+test('stubbing', async (context) => {
```
sxzz: It's ok to remove this and keep multiple entries

tests/index.test.ts
```diff
@@ -256,3 +256,32 @@ test('minify', async (context) => {
...
+  const { snapshot } = await testBuild(context, files, {
+    entry: 'index.ts',
+    stub: true,
+  })
+  expect(snapshot).contains(
+    'export * from "[CWD]/tests/temp/stubbing/index.ts"',
+  )
+})
+
+test('stubbing with multiple entries', async (context) => {
```
sxzz:
```suggestion
test('stub', async (context) => {
```

src/features/stubbing.ts
```diff
@@ -0,0 +1,41 @@
...
+      /**
+       * Generate a stub to each entry point.
+       */
+      let code = ''
+      // Entry is always a Record<string, string>
+      if (typeof entry === 'object') {
+        // eslint-disable-next-line unused-imports/no-unused-vars
+        for (const [_, file] of Object.entries(entry)) {
+          const fileName = path.basename(file)
+          code += `export * from "${path.resolve(fileName)}";\n`
```
sxzz: Replace `\` with `/` or it will break Windows

Gugustinette: Were you talking about the path.resolve ? Or \n ? Or both ?

src/index.ts
Gugustinette: I turned off the plugins you mentioned, but I must say I'm not sure which one should be disabled or not 😭

src/features/stubbing.ts
Gugustinette: So [`load`](https://rollupjs.org/plugin-development/#load) instead of [`generateBundle`](https://rollupjs.org/plugin-development/#generatebundle) ?

I tried using `load` but did not manage to use it correctly to get the same behavior (again, I'm new to bundlers sorry 😭)

Gugustinette:
I made an example repo to test the feature and compare it to the unbuild stub : https://github.com/Gugustinette/tsdown-stub-test

Of course, for now it does not work, we get this output using unbuild :
```js
import { createJiti } from "../../../node_modules/jiti/lib/jiti.mjs";

const jiti = createJiti(import.meta.url, {
  "interopDefault": true,
  "alias": {
    "@tsdown-stub-test/foo": "/Users/augustinmercier/Desktop/tsdown-stub-test/packages/foo"
  },
  "transformOptions": {
    "babel": {
      "plugins": []
    }
  }
})

/** @type {import("/Users/augustinmercier/Desktop/tsdown-stub-test/packages/foo/src/index")} */
const _module = await jiti.import("/Users/augustinmercier/Desktop/tsdown-stub-test/packages/foo/src/index.ts");

export const foo = _module.foo;
```

And this output using tsdown :
```js
export * from "/Users/augustinmercier/Desktop/tsdown-stub-test/packages/foo/src/index.ts";
export { default } from "/Users/augustinmercier/Desktop/tsdown-stub-test/packages/foo/src/index.ts";
```

Gugustinette:
@sxzz Do we copy the unbuild behavior by using [`jiti`](https://github.com/unjs/jiti) ?

Only drawback I see is that it implies adding jiti as a dependency just for the stub feature, and it is [2Mb](https://www.npmjs.com/package/jiti). But jiti has 0 dependency so that's not too bad I guess ?

I thought about re-creating the jiti behavior ourself, but that might be too much work ?

Curious about how you see things on this side 🤔

lishaduck:
> Do we copy the unbuild behavior by using [jiti](https://github.com/unjs/jiti) ?

I might consider oxc-node, which is smaller and more consistent (because we use oxc). I think there's an API we could use?

lishaduck:
Yeah, I think either oxc-node or unloader would be best, _if_ we can get either of them to expose a jiti-like API. It looks right now, they only support being used as a loader hook.

Gugustinette:
Well it looks like @sxzz [used `unloader` on his demo](https://github.com/sxzz/stub-demo/blob/main/unloader.config.ts), so I guess this is what we are going for.

I did not pay intention to this part of his demo, that's my bad 😭

lishaduck:
> Well it looks like @sxzz [used `unloader` on his demo](https://github.com/sxzz/stub-demo/blob/main/unloader.config.ts), so I guess this is what we are going for.

Yeah, that's what I'd initially meant when I filed the issue, but upon actually looking at it, I don't think unloader offers an importing API (neither does oxnode, but it looks like it'd be pretty easy to work around that 🤷‍♂️).
Then again, it might just be better to make an importing API for unloader, as they're both Kevin's (assuming you approve, @sxzz).

Gugustinette:
@sxzz I made some experimental work here : https://github.com/Gugustinette/jitdown

Gugustinette:
Great investigations !
I agree ; should we add something in the documentation, stating tsdown won't support stub mode (atleast for now) and recommand using watch mode ? Maybe referencing a link to [your comment](https://github.com/rolldown/tsdown/pull/164#issuecomment-2849720617) also.

sxzz:
Sure, could you please send a PR for it?

Gugustinette:
Yes !

Where in the doc is that most relevant ? I thought about either :
- A tip in the watch mode doc
- A dedicated page in "Advanced", if we want more advanced explanation about what stub is, why It won't be supported,...
- Adding Q&A section ? I did this in Nuxt Leaflet and I feel it can be useful for common questions, instead of searching through issues (as the Q&A is searchable through the search bar) : https://leaflet.nuxtjs.org/about/q&a.html

sxzz:
Yeah, we should has a FAQ (Frequently Asked Questions) page, which is a good idea.

PR #166, opened 2025-04-28
feat(workspace): add workspace config file resolver

fu050409:


src/options.ts
```diff
@@ -46,10 +47,37 @@ export interface MinifyOptions {
...
 /**
  * Options for tsdown.
  */
 export interface Options {
+  /// workspace options
+  /**
+   * Workspace configs.
+   * @default false
+   */
+  workspace?: Workspace | false
```
sxzz: Remove `workspace` from `Options`, and have a new function `defineWorkspace`

src/options.ts
```diff
@@ -46,10 +47,37 @@ export interface MinifyOptions {
...
 
+/**
+ * Options for tsdown workspace.
+ */
+export interface Workspace {
+  /**
+   * Workspace members.
+   * @default 'auto'
+   */
+  members?: string[] | 'auto'
```
```suggestion
  packages?: string[] | 'auto'
```

fu050409: All right

src/options.ts
fu050409: Well.

The reason for this is that it was considered to allow users to create content in the root directory as well, but this does indeed increase complexity and is not convenient. As you wish, I will modify it.

sxzz:
Thank you for your contribution! I implemented this feature in commit 6f9629c5220c2c7cb608bfa1cd5c0568f2f9c696 and incorporated some of your code, so I have also added you as a co-author.

johnpyp:
Hi, there are no docs for this feature and it's unclear how it works and what it does / doesn't do, that would be greatly appreciated.

sxzz:
Yes, it's still an experimental feature, so there is currently no documentation available. I'll look into it further and write some documentation.

Issue #176, opened 2025-04-29
`node:` protocol should be removed when target doesn't support it


SuperchupuDev:
### Describe the bug

running tsdown doesn't remove the `node:` protocol from imports on old enough targets, like `node12`. this makes running the bundled package potentially incompatible in those versions (the node protocol wasn't fully available in node 12)

### Reproduction

https://github.com/SuperchupuDev/tsdown-node-protocol-repro

### System Info

```Shell
System:
    OS: Linux 6.14 EndeavourOS
    CPU: (16) x64 AMD Ryzen 7 8845HS w/ Radeon 780M Graphics
    Memory: 19.24 GB / 30.65 GB
    Container: Yes
    Shell: 5.2.37 - /bin/bash
  Binaries:
    Node: 23.11.0 - ~/.nvm/versions/node/v23.11.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v23.11.0/bin/npm
    pnpm: 10.8.1 - ~/.nvm/versions/node/v23.11.0/bin/pnpm
```

### Used Package Manager

pnpm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/sxzz/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/sxzz/contribute).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [x] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

MatteoGabriele:
Sorry for my questions, but isn't Node v12 a bit outdated since EOL was back in 2022?

SuperchupuDev:
yes, sadly i can't bump the minimum required version because many libraries that depend on mine are node12 compatible

sxzz:
It's hard to manage which Node.js version is be remove `node:` protocol. There's an option `removeNodeProtocol` in tsup, maybe we can implement.

sxzz:
Added `removeNodeProtocol ` in #179

PR #179, opened 2025-04-30
feat(node-protocol): add option to remove the `node:` protocol

wChenonly:
### Description
fix:https://github.com/rolldown/tsdown/issues/176

@sxzz the script needs to update the document,，
In addition, I created a new plugins file, and I'm going to move it to plugins in the next pr, which will be clearer. What do you think?
image

### DIFF
--- a/src/features/node-protocol.ts
+++ b/src/features/node-protocol.ts
@@ -1,12 +1,12 @@
 import type { Plugin } from 'rolldown'
 
 /**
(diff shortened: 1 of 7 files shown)


sxzz:
I think `features` folder is enough.

wChenonly:
> I think `features` folder is enough.

get，move

sxzz (approved):

PR #180, opened 2025-04-30
chore: Import plugins into a single file and correct comment examples

wChenonly:
### Description
Import plugins into a single file and correct comment examples

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -184,7 +184,7 @@ export interface Options {
    *
    * @example
    * // With removeNodeProtocol enabled:
-   * import('node:fs'); // becomes import('fs')
+   * import('node:fs'); // becomes import('node')
    */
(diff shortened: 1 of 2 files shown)


src/options.ts
```diff
@@ -184,7 +184,7 @@ export interface Options {
    *
    * @example
    * // With removeNodeProtocol enabled:
-   * import('node:fs'); // becomes import('fs')
+   * import('node:fs'); // becomes import('node')
```
lishaduck, line 187: That's wrong, removing the `node:` prefix from `node:fs` should give `fs`. If it doesn't work that way, it's a bug.

sxzz:
`plugins.ts` entry is only for exposing plugins that can be used by end-users. and JSDoc of `removeNodeProtocol` is correct.

PR #194, opened 2025-05-03
fix: outDir cannot be an empty string

btea:
### Description
If outDir is an empty string, the current logic will clear all files in the root directory of the current project, resulting in a build error.
https://github.com/rolldown/tsdown/blob/3630a61e7c961370b46d71861fceb2f0ea704b4d/src/index.ts#L64-L66

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 import path from 'node:path'
 import process from 'node:process'
 import { fileURLToPath } from 'node:url'
-import { blue, green } from 'ansis'
+import { green } from 'ansis'
 import Debug from 'debug'
 import {
   build as rolldownBuild,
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -55,6 +55,16 @@ export async function build(userOptions: Options = {}): Promise<void> {
     debug('No config file found')
   }
 
+  for (const config of configs) {
+    if (config.outDir === process.cwd()) {
```
sxzz: Put them into `resolveOptions`

btea: Done.

src/options.ts
```diff
@@ -321,6 +321,13 @@ export async function resolveOptions(options: Options): Promise<{
       } = subOptions
 
       outDir = path.resolve(outDir)
+      if (outDir === process.cwd()) {
+        throw new Error(
+          'Output directory cannot be the same as the current working directory. ' +
+            `Please specify a different watch directory using ${blue`watch`} option,` +
```
sxzz: Only throw when `clean` or `watch` is enabled, also error message should be updated.

btea: The error message has been adjusted according to the previous one, but I’m unsure if it’s appropriate.

sxzz (approved):

PR #199, opened 2025-05-05
fix: `clean: true` not work in Windows

northword:
### Description

When `clean` is the default (`true`), it will be assigned with `dist`, whereas in the current behavior, `dist` is resolved to an absolute path before it is assigned to `clean`, and the `glob` is set to `cwd` in `cleanOutDir`, which results in the `glob` getting an empty list of paths, leading to no files being cleaned.

### DIFF
--- a/src/features/clean.ts
+++ b/src/features/clean.ts
@@ -1,6 +1,7 @@
 import Debug from 'debug'
 import { glob } from 'tinyglobby'
 import { fsRemove } from '../utils/fs'
+import { slash } from '../utils/general'
 import { logger } from '../utils/logger'
 import type { Options, ResolvedOptions } from '../options'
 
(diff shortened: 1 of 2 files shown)


sxzz:
Are you using Windows? Try `clean = [outDir.replaceAll('\\', '/')]` in `src/features/clean.ts`, and it may resolve the issue.

northword:
> Are you using Windows? Try `clean = [outDir.replaceAll('\\', '/')]` in `src/features/clean.ts`, and it may resolve the issue.

Yes, thanks for pointing out that the `glob` schema not being able to contain `\\` was the crux of the issue, PR has been updated.

src/features/clean.ts
```diff
@@ -13,7 +14,7 @@ export async function cleanOutDir(configs: ResolvedOptions[]): Promise<void> {
 
   for (const config of configs) {
     if (!config.clean.length) continue
-    const files = await glob(config.clean, {
+    const files = await glob(config.clean.map(slash), {
```
sxzz: It should be normalized in `resolveClean` fn, instead of here.

northword: Updated.

sxzz (approved):

PR #227, opened 2025-05-13
fix(pnp): resolve typescript peer dependency with Yarn PnP

gronxb:
### Description

When you run the following command in a Yarn PnP environment:
```sh
yarn tsdown index.ts --dts
```

we get the following error:
image

In a Yarn PnP environment, using the `--dts` flag during build causes a peer dependency error. This is required by `rolldown-plugin-dts`.

Although `typescript` is specified as a peer dependency in `rolldown-plugin-dts`, it's bundled within `tsdown`, so if you use only `tsdown`, the peer dependency is not propagated. Therefore, `tsdown` also needs to declare `typescript` as a peer dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,8 @@
   "peerDependencies": {
     "publint": "^0.3.0",
     "unplugin-lightningcss": "^0.4.0",
-    "unplugin-unused": "^0.5.0"
+    "unplugin-unused": "^0.5.0",
+    "typescript": "^5.0.0"
   },
   "peerDependenciesMeta": {
     "publint": {
@@ -70,6 +71,9 @@
       "optional": true
     },
     "unplugin-unused": {
+      "optional": true
+    },
+    "typescript": {
       "optional": true
     }
   },
--- a/package.json
+++ b/package.json
@@ -59,21 +59,21 @@
   },
   "peerDependencies": {
     "publint": "^0.3.0",
+    "typescript": "^5.0.0",
     "unplugin-lightningcss": "^0.4.0",
-    "unplugin-unused": "^0.5.0",
-    "typescript": "^5.0.0"
+    "unplugin-unused": "^0.5.0"
   },
   "peerDependenciesMeta": {
     "publint": {
+      "optional": true
+    },
+    "typescript": {
       "optional": true
     },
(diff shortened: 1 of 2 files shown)


sxzz:
`rolldown-plugin-dts` is declared in `dependencies` field, not bundled.

gronxb:
Oh, my mistake.

> it’s bundled within tsdown,

 this part should be rephrased to say that tsdown includes it as a dependency.

gronxb:
It is still valid that the peer of rolldown-plugins-dts needs to propagate down to tsdown.

sxzz:
This issue is not specific to Yarn PnP; it can also be reproduced with npm. Installing `typescript` in the project resolves the problem. We should also add a note about this in the documentation.

```
ERROR  Error: Cannot find module 'typescript'
```

gronxb:
@sxzz 

This issue can be reliably reproduced in a Yarn workspace.

Please check the following repository:  
https://github.com/gronxb/yarn-pnp-tsdown-repro

To reproduce:

```bash
git clone https://github.com/gronxb/yarn-pnp-tsdown-repro 
cd yarn-pnp-tsdown-repro
yarn
cd a-package
yarn tsdown index.ts --dts
```


Here is the package.json for reference:
```
{
  "name": "a-package",
  "packageManager": "yarn@4.8.0",
  "devDependencies": {
    "tsdown": "^0.11.5",
    "typescript": "^5.8.3"
  }
}
```

gronxb:
You can resolve the issue by properly configuring the peer of tsdown in `.yarnrc.yml` as follows:
```yml
packageExtensions:
  "tsdown@*":
    peerDependencies:
      typescript: "^5.0.0"
```

Since `tsdown` also declares typescript as a peer, I believe the issue can be resolved only if the peer dependency of `rolldown-plugin-dts` is correctly propagated to the user’s project through `tsdown`.

sxzz (approved):
Thanks, merged!

gronxb:
switched from using tsup to tsdown, even though it wasn’t directly related to the issue, and it seems to have much better compatibility for dual ESM and CJS packages.
Thank you for your work—it’s an awesome project. 🥰

PR #233, opened 2025-05-15
feat: add runtime assistant check plugin to support target environment

wChenonly:
### Description
support better tips. The error message printed by the previous upper tool is not friendly, and the plug-in is temporarily used to handle it in tsdown.

In the future,  rolldown to support more friendly printing.

before 
image

after
image


src/index.ts
```diff
@@ -225,6 +226,7 @@ async function getBuildOptions(
       const { Unused } = await import('unplugin-unused')
       plugins.push(Unused.rolldown(unused === true ? {} : unused))
     }
+
```
sxzz: Remove the blank line

src/features/runtime-helper-check.ts
```diff
@@ -0,0 +1,30 @@
+import { logger } from '../utils/logger'
+import type { Plugin } from 'rolldown'
+
+let isRuntimeMissing = false
+
+export function RuntimeHelperCheckPlugin(target: string[]): Plugin {
+  return {
+    name: 'tsdown:runtime-helper-check',
+    resolveId(source: string) {
```
sxzz: make it to `resolveId.handler` and add `resolveId.filter`, see https://rolldown.rs/guide/plugin-development#plugin-hook-filters

sxzz:
```suggestion
    resolveId(id: string) {
```

sxzz (approved):

PR #244, opened 2025-05-19
fix: 🚑 fix failing docs build

okineadev:
### Description

Fixed documentation build errors noted in https://app.netlify.com/projects/tsdown/deploys/682b3101d65d1600086495fb

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,6 @@
     "docs/**/config-options.md": true,
     "docs/**/workspace.md": true,
     "docs/**/type-aliases/*.md": true
-  }
+  },
+  "typescript.tsdk": "node_modules/typescript/lib"
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -83,8 +83,8 @@ catalogs:
       specifier: ^1.5.5
       version: 1.5.5
     vitepress-plugin-llms:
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -5,5 +5,6 @@
     "docs/**/config-options.md": true,
     "docs/**/workspace.md": true,
     "docs/**/type-aliases/*.md": true
-  }
+  },
+  "typescript.tsdk": "node_modules/typescript/lib"
```
okineadev, line 9: This is necessary so that VS Code uses the TS version to find errors from `node_modules` instead of the VS Code TS version, this saves a lot of time on debugging, because sometimes there are errors due to differences in TS versions, for example when VS Code says that there is an error, but `tsc` in the terminal does not have it

sxzz:
Actually, simply enabling `enforce: 'post'` can resolve this issue. Do you think it would be better to address this on the vitepress-plugin-llms side?

okineadev:
> Actually, simply enabling `enforce: 'post'` can resolve this issue. Do you think it would be better to address this on the vitepress-plugin-llms side?

I'll check it later.

okineadev:
@sxzz I think you can merge this PR now

sxzz (approved):

okineadev:
yay!

PR #251, opened 2025-05-21
feat: print tsdown version

btea:
### Description
Printing the currently running version helps verify and compare different versions when switching between them.

image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,6 +11,7 @@ import {
   type RolldownPluginOption,
 } from 'rolldown'
 import { exec } from 'tinyexec'
+import { version } from '../package.json'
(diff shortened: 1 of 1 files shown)


sxzz:
Already printed
https://github.com/rolldown/tsdown/blob/8dfcd2ae3c378807b88fd7e8c474a782e83479c5/src/cli.ts#L61

btea:
Sorry, didn't notice it was printed at the beginning.

Do you think it is necessary to print the version information when executing the `migrate` command?

sxzz:
It's generally not necessary to worry about tsdown's version during migration. However, if you do need to check the version, simply run `tsdown -v`.

btea:
Got it.

PR #307, opened 2025-06-08
fix: add "types" field to subpath exports map generation

### LINKED ISSUE #253: when `exports: true`, `subExport` doesn't include type definition

no reproduction.

### What is expected?

Let's imagine you create an export map for unplugin.
I'd like to have the exports map with type definition
```sh
tsdown --dts --exports
```
generates export map like this
```json
exports: {
    './': {
      typs: './dist/index.d.ts',
     default: './dist/index.js'
   }
}
```

or 

```json
exports: {
    './': {
       import: {
         types: './dist/index.d.js',
         default: './dist/index.d.js'
       },
       require: {
         types: './dist/index.d.cjs',
         default: './dist/index.cjs'
      }
   }
}
```

In v0.12.1
```sh
tsdown --dts --exports
```
generates export map like this
```json
exports: {
    './': './dist/index.js'
}
```

### Any additional comments?

Also I'd like to have conditional export for deno or bun
https://bun.sh/docs/runtime/modules#importing-packages

but this can be a future work.

YangJonghun:
### Description
When developers generate a bundle with multiple entry points, tsdown currently does not automatically generate the "types" field for subpath exports.

For example, if a developer provides the following entries:
```
["src/index.ts", "src/foo.ts"]
```

tsdown should generate the following exports map:
```diff
"exports": {
  ".": {
    "import": {
      "types": "./index.d.ts",
      "default": "./index.js"
    },
    "require": {
      "types": "./index.d.cts",
      "default": "./index.cjs"
    }
  },
  "./foo": {
    "import": {
      "types": "./foo.d.ts",
      "default": "./foo.js"
    },
    "require": {
      "types": "./foo.d.cts",
      "default": "./foo.cjs"
    }
  },
  "./package.json": "./package.json",
},
"main": "./index.cjs",
"module": "./index.js",
"types": "./index.d.cts",
```

### Linked Issues

closes #253


src/features/exports.ts
```diff
@@ -204,6 +208,7 @@ function genSubExport(
     }
     if (es) value[dualFormat ? 'import' : 'default'] = es
     if (cjs) value[dualFormat ? 'require' : 'default'] = cjs
+    if (cjsTypes || esmTypes) value.types = cjsTypes || esmTypes
```
sxzz: `types` should precede `import` or `require`.

src/features/exports.ts
```diff
@@ -130,6 +132,8 @@ export async function generateExports(
         if (chunk.facadeModuleId && !subExport.src) {
           subExport.src = `./${slash(path.relative(pkgRoot, chunk.facadeModuleId))}`
         }
+      } else if (!isIndex && (format === 'cjs' || format === 'es')) {
```
sxzz: Why only add `types` for non-index?

sxzz:
I think we'd better to have an option for this `types` generation, see https://github.com/rolldown/tsdown/issues/253#issuecomment-2900595087

YangJonghun:
@sxzz 
Thank you for quick review. I've updated the code to separate the `require` and `import` cases, and reorder the "types" field to be declared earlier.

src/features/exports.test.ts
```diff
@@ -133,12 +133,18 @@ describe.concurrent('generateExports', () => {
...
-    await expect(results).resolves.toMatchInlineSnapshot(`
+    expect(results).toMatchInlineSnapshot(`
       {
         "exports": {
           ".": {
-            "import": "./foo.js",
-            "require": "./foo.cjs",
+            "import": {
+              "default": "./foo.js",
+              "types": "./foo.d.ts",
```
sxzz, line 167: Still incorrect

YangJonghun: Since writeFile also uses JSON.stringify, key order validation is handled in the following line:
expect(JSON.stringify(results.exports, undefined, 2)).toMatchInlineSnapshot(...).

src/features/exports.test.ts
```diff
@@ -123,57 +123,91 @@ describe.concurrent('generateExports', () => {
...
+        "module": "./foo.js",
+        "types": "./foo.d.ts",
         "exports": {
-          ".": {
-            "import": "./foo.js",
-            "require": "./foo.cjs",
-          },
-          "./package.json": "./package.json",
-        },
+          ".": "./foo.js",
```
sxzz, line 141: I just added `dts with single entry` test, and `types` field should be added here.

YangJonghun: Wouldn't it make sense for the "types" setting to affect the [root-level types](https://github.com/rolldown/tsdown/blob/main/src/features/exports.ts#L182) too?

sxzz: In fact, setting `exports.types` itself doesn't make much difference, since TypeScript will look for `.d.ts` files based on the `.js` entry. However, explicitly specifying it may improve resolution speed. Regardless, since we have this option, we should set it for the main entry as well to maintain consistency.

YangJonghun: Thanks for your explanation.
If there's anything else I need to change, please feel free to let me know.

sxzz: I mean `types` field should be added here.

osnoser1-slalom:
I'm waiting this feature, it is really helpful, thank you for this work @YangJonghun and @sxzz

jadejr:
Will it be easy to add custom conditions to this later (not necessarily in this PR)? That's what I (currently) use to get live types. I've been managing it manually, but it'd be cool if tsdown could do it for me.

sxzz:
Closing this as stale, and `types` field can be omitted, and customized https://tsdown.dev/options/package-exports#customizing-exports

PR #320, opened 2025-06-13
fix: regexp needs to reset lastIndex each time test is called

btea:
### Description
If the regular expression passed in carries the `g` modifier, multiple tests may produce unexpected results.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

### DIFF
--- a/src/features/external.ts
+++ b/src/features/external.ts
@@ -22,7 +22,11 @@ export function ExternalPlugin(options: ResolvedOptions): Plugin {
         const noExternalPatterns = toArray(noExternal)
         if (
           noExternalPatterns.some((pattern) => {
-            return pattern instanceof RegExp ? pattern.test(id) : id === pattern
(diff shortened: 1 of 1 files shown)


src/features/external.ts
```diff
@@ -22,7 +22,11 @@ export function ExternalPlugin(options: ResolvedOptions): Plugin {
         const noExternalPatterns = toArray(noExternal)
         if (
           noExternalPatterns.some((pattern) => {
-            return pattern instanceof RegExp ? pattern.test(id) : id === pattern
+            if (pattern instanceof RegExp) {
+              pattern.lastIndex = 0
+              return pattern.test(id)
```
sxzz, line 27: `pattern.test` won't change `lastIndex` property but `exec` does. It should handle by user-land I think.

sxzz (approved):

PR #323, opened 2025-06-13
fix: multiple config watch files affect each other

### LINKED ISSUE #315: Infinite rebuild loop in watch mode when `outDir` is not `dist` in multi-config setup

Reproduction: https://bolt.new/~/github-v69gmpbw

Steps to reproduce manually:
1. Create a configuration file with two objects, where one of them has an `outDir` option not set to `dist`. For example:
```ts
export default defineConfig([
  { entry: './src/index.ts' },
  { entry: './src/cli.ts', outDir: 'bin' }
])
```
2. Run `pnpm tsdown -w`
3. Change some source code in `index.ts` or `cli.ts`

### What is expected?

Only one rebuild should be triggered.

The rebuild process enters an infinite loop.

### Any additional comments?

Thanks a lot for this awesome tool! It's exactly what I’ve been looking for to build TS libs with minimal config. Really happy I discovered it.

btea:
### Description
fix #315

### DIFF
--- a/src/features/watch.ts
+++ b/src/features/watch.ts
@@ -11,6 +11,7 @@ export async function watchBuild(
   configFiles: string[],
   rebuild: () => void,
   restart: () => void,
+  ignoreWatch: string[] = [],
 ): Promise<FSWatcher> {
   if (typeof options.watch === 'boolean' && options.outDir === options.cwd) {
     throw new Error(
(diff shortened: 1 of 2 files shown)


sxzz:
Thank you for your PR! However, I'm currently refactoring the watcher to use Rolldown itself, which is smarter and doesn't require specifying a path to watch. #329

Issue #288, opened 2025-06-01
Add `banner` and `footer` options for output file customization


junwen-k:
## Description
It would be useful to have `banner` and `footer` configuration options similar to `tsup` (`esbuild`) in `tsdown` to allow inserting custom content (like comments or notices) at the beginning and end of output files.

## Use Cases
1. Adding warning notices about generated files
2. Including license information
3. Adding build metadata or timestamps
4. Inserting environment-specific configuration

## Related
- Similar feature in tsup: https://github.com/egoist/tsup/issues/451

## Proposed API

Similar to `tsup` / `esbuild`:

```typescript
export default defineConfig({
  // ... other config
  banner: {
    js: `/**
     * ⚠️  WARNING: This is a generated file
     * Do not edit this file directly.
     * Generated: ${new Date().toISOString()}
     */`,
  },
  footer: {
    js: `// End of generated file`,
  },
})
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/sxzz/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/sxzz/contribute).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

kricsleo:
I think it's already supported by `tsdown` (`rolldown`)

```ts
import { defineConfig } from 'tsdown'

export default defineConfig({
  outputOptions: {
    banner: '// hello banner',
    footer: '// hello footer',
  },
})
```

Output:

```js
// hello banner
//#region src/index.ts
const a = 42;

//#endregion
export { a };
// hello footer
```

TheAlexLichter:
Indeed - [Rolldown implemented it](https://bundler.sxzz.dev/#eNptjkELgkAQhf/KMJcuohCdNjoYGHVRMI8eNN0NQ2ZlnSVB9r+3JhRBMIf5+B6PN6NCMWNHrZxCHpefUHw5wMajnAZtGBpNI4PSGg6w23onUbCx0gVodN+3+kmhz6juHj4+VX/MT2krVW17hrkkAG15sCxWALjVRNII2EQRHOM0TfL9JliVX8GLgsq74ny5gr8YTllWJHn1DrmSHDrnXvUuSCw=).

Might be worth to add it to the docs as info (possibly in the migration guide?)

sxzz:
It's a little different from tsup's

Eptagone:
[...]

PR #338, opened 2025-06-21
feat: add banner, footer options

gwansikk:
### Description

Hello. I'm an enthusiast of tsup and have great interest in the tsdown project.
This PR is merely a suggestion and I submitted it with the intention of having a discussion with you.
If this PR doesn't align with the direction of tsdown, please feel free to close it.

tsdown is a project inspired by tsup, and in fact, many of its configuration structures and core philosophies are designed similarly.
For that reason, tsup users who switch to tsdown inevitably expect a similar developer experience (DX).
I believe the banner and footer features fall within that context. These are widely used in various bundlers such as Rollup, esbuild, and webpack, and have been supported in tsup as well.

However, since tsdown currently uses rolldown internally, although the functionality itself is easy to implement, the configuration method has changed.
As a result, existing tsup users face an unnecessary entry barrier by having to look up documentation again or modify their existing configurations.
I believe this isn't just an interface difference, but a factor that causes friction during migration.

To become a project that tsup users can naturally transition to, I think it's important to maintain as much functional parity and consistency in configuration interface as possible. (Just my opinion.)
This PR, as part of that, proposes adopting the same banner/footer configuration style as tsup.

Of course, due to rolldown's structure, the future direction may change.
If so, I think we could proceed by separating it through a breaking change via versioning later on.
However, at this stage, encouraging user adoption with low friction and then shaping the direction seems like a more strategic approach.
The fact that users can transition with a similar API and minimal configuration changes could offer great convenience.

### Linked Issues

closed: #288

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,8 +16,8 @@ catalogs:
       specifier: ^2.2.3
       version: 2.2.3
     '@sxzz/test-utils':
-      specifier: ^0.5.6
-      version: 0.5.6
+      specifier: ^0.5.7
+      version: 0.5.7
     '@types/debug':
       specifier: ^4.1.12
       version: 4.1.12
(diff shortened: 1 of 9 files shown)


gwansikk:
@sxzz I want to hear your opinion and continue the task based on it (Please share your thoughts!)

sxzz (approved):

Issue #339, opened 2025-06-21
[Exports] Allow excluding files via regex or glob


TheAlexLichter:
### Clear and concise description of the problem

When writing exports, I want to exclude files based on regex or globs that shouldn't be exported.

### Suggested solution

An `exclude` option under `ExportOptions` that takes an array of regexes or globs that will be honored

### Alternative

Writing a custom exports fn, but that seems to be quite brittle.

### Additional context

https://x.com/colinhacks/status/1936539368573083894

### Validations

- [x] Follow our [Code of Conduct](https://github.com/sxzz/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/sxzz/contribute).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sxzz:
I don't really understand why this feature is needed. tsdown only adds entry points to exports. Why would a user add a file to entrypoints but not want to expose it? Even if that's the case, is this a common use case? Isn't customExports already sufficient for this?

TheAlexLichter:
> I don't really understand why this feature is needed. tsdown only adds entry points to exports. Why would a user add a file to entrypoints but not want to expose it?

If think this might be more interesting if tsdown would also allow *glob* entry points. Otherwise it isn't helpful at the moment, I agree. 

> Even if that's the case, is this a common use case? Isn't customExports already sufficient for this?

With globs at least more likely. `customExports` is fine and works already for it but might be a bit verbose depending on the requirements.

sxzz:
I'm not sure it's worth introducing the `picomatch` dependency for this uncertain feature.

TheAlexLichter:
> I'm not sure it's worth introducing the `picomatch` dependency for this uncertain feature.

Fair. Maybe let's discuss #341 at first?

sxzz:
Merged `exports.exclude`, but without glob support. Only exact matching or RegExp patterns are supported.

sxzz:
I finally added glob support back 84b68bb6aa3e1834b230750c645aedca2973c0a1 because `picomatch` was already added as a dependency in bcb7a62fa65b38ddb29a8a947475207f870d7cc4. More importantly, since `tinyglobby` uses `picomatch` under the hood, we haven’t introduced any new dependencies.

PR #340, opened 2025-06-21
feat: export exclude option

TheAlexLichter:
Resolves https://github.com/rolldown/tsdown/issues/339

### DIFF
--- a/src/features/pkg/exports.ts
+++ b/src/features/pkg/exports.ts
@@ -26,7 +26,7 @@ export interface ExportsOptions {
   all?: boolean
 
   /**
-   * Define filenames or RegExps to exclude files from exports.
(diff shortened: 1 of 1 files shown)


smorimoto:
Any updates on this?

TheAlexLichter:
@smorimoto this needs more discussion I think

sxzz:
@TheAlexLichter I finally added glob support back 84b68bb6aa3e1834b230750c645aedca2973c0a1 because `picomatch` was already added as a dependency in bcb7a62fa65b38ddb29a8a947475207f870d7cc4. More importantly, since `tinyglobby` uses `picomatch` under the hood, we haven’t introduced any new dependencies.

PR #356, opened 2025-06-28
feat: add `watch.usePolling` , `watch.atomic` and `watch.path`

### LINKED ISSUE #355: `watch.usePolling` option similar to vite's, to use a polling stratgey to decide when to rebundle

tsdown is currently missing `watch.usePolling` option similar to vite's `server.watch.usePolling` which uses polling to detect if needs to hot-reload.

### Suggested solution

currently we have 

```ts
{
  watch?: boolean | string | string[]
}
```

Ideally we should also have

```ts
{ 
  watch?: {
    path?: string| string[], // defaults to watch all
    usePolling?: boolean = false,
    useFsEvents?: boolean = true,
  }
}
```

### Alternative

_No response_

### Additional context

This can be useful in certain situations as using Docker with WSL2 where file system events are not propagated correctly.

### Validations

amrbashir:
### Description

Adds support for chokidar `usePolling` and `atomic` options similar to vite's

### Linked Issues

closes #355

### Additional context

I see https://github.com/rolldown/tsdown/pull/329 which removes chokidar in favor of rolldown's watcher, which afaik doesn't have polling option (correct me if I am wrong), so if this PR needs to be closed and the feature should be implemented in rolldown instead, I don't mind, I can contribute to rolldown.

### DIFF
--- a/src/features/watch.ts
+++ b/src/features/watch.ts
@@ -21,7 +21,13 @@ export async function watchBuild(
   }
 
   const files = toArray(
-    typeof options.watch === 'boolean' ? options.cwd : options.watch,
+    typeof options.watch === 'boolean'
+      ? options.cwd
+      : typeof options.watch === 'object' && 'path' in options.watch
+        ? options.watch.path
+        : typeof options.watch === 'object'
+          ? options.cwd
+          : options.watch,
   )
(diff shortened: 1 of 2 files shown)


sxzz:
I'm refactoring the watcher using native Rolldown. See #329.

PR #376, opened 2025-07-15
fix(workspace): path compatibility issue under windows

RSS1102:
### Description

In Windows, the package index package will have a path error, I use [normalizepath ](https://vite.dev/guide/api-javascript.html#normalizepath) to normalize the path to solve this compatibility error.

559f8f3801df5308601a4a4eeb3a0aea

### DIFF
--- a/src/options/index.ts
+++ b/src/options/index.ts
@@ -3,6 +3,7 @@ import process from 'node:process'
 import { blue } from 'ansis'
 import Debug from 'debug'
 import { glob } from 'tinyglobby'
+import { normalizePath } from 'vite'
 import { resolveClean } from '../features/clean'
 import { resolveEntry } from '../features/entry'
 import { resolveTarget } from '../features/target'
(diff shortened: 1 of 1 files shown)


sxzz (approved):

src/options/index.ts
```diff
@@ -3,6 +3,7 @@ import process from 'node:process'
 import { blue } from 'ansis'
 import Debug from 'debug'
 import { glob } from 'tinyglobby'
+import { normalizePath } from 'vite'
```
sxzz: Vite is not a production dependency of tsdown.

sxzz:
Use `slash` from `src/utils/general.ts` instead.

RSS1102:
I also had no success building tsdown on Windows, waiting me test of the fixed functionality using pkg-pr-new.

RSS1102:
> 改用`slash`from `src/utils/general.ts`。

ok

RSS1102:
It seems that there are many places in the project where `path.resolve` is used. Will they all have Windows compatibility error?

sxzz:
Some usages are intended solely for display or file system operations, not for searching or matching.

RSS1102:
Ok, let me try it out with pkg.pr.new to my project.

PR #381, opened 2025-07-18
fix(types): keep type of options

posva:
### Description

When using `defineConfig({})`, the return value cannot be spread because the type includes a function. Since the `defineConfig()` is just the identity function, it might be better to keep the type as is. Another option is to create an overload function

### Additional context

The current workaround is to use `satisfies UserOptions`

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -3,9 +3,9 @@ import type { UserConfig, UserConfigFn } from './options'
 /**
  * Defines the configuration for tsdown.
  */
-export function defineConfig(
-  options: UserConfig | UserConfigFn,
-): UserConfig | UserConfigFn {
+export function defineConfig<const T extends UserConfig | UserConfigFn>(
+  options: T,
+): T {
   return options
 }
 


sxzz (approved):

sxzz:
Thanks bro!

sxzz:
@posva I reverted this PR via 7c32a1d8a177aee1a5d2e99f4a1ba400b9b00ff3 because:

- **Overloading `defineConfig`** (similar to how Vite handles it):
```ts
export function defineConfig(options: UserConfig): UserConfig;
export function defineConfig(options: UserConfigFn): UserConfigFn;
export function defineConfig(
  options: UserConfig | UserConfigFn,
): UserConfig | UserConfigFn;
```

- **Using generics**:
  - Provides a more direct error message.
  - However, it doesn't allow jumping to the definition of properties by clicking, and no auto suggestion/completion.

- **Using a union**:
  - Makes it impossible to extract specific types.

After evaluating these options, I decided to go with overloading, as Vite does. If you need to achieve type inference, using `satisfies` is a more appropriate solution.

posva:
Of course, I didn’t expect it to break jump to definition. TIL

PR #403, opened 2025-08-01
feat: warn user if config includes cjs format

### LINKED ISSUE #260: Info/Warn when CJS is used

`tsdown` supports CJS as a possible output format. Given that [`require(esm)`](https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js/) exist nowadays and Node 18 is EOL, ESM or at least dual export should be preferred and CJS should be discouraged.

### Suggested solution

While there are many ways to nudge towards using ESM, from being radical and *remove* CJS support over to more gentle ways, emitting an info/warning when using CJS could be helpful and make people aware of the current situation where switching is less costly than before, ideally paired with some blog post or documentation (could be living directly in the `tsdown` docs).

### Alternative

_No response_

### Additional context

_No response_

### Validations

Gugustinette:
### Description

Inform the user when using the `cjs` format.

### Linked Issues

Resolves #260 

### Additional context

- Not sure about the phrasing I used
- I added a link to [this paragraph from the Node.js website](https://nodejs.org/fr/learn/modules/publishing-a-package#how-did-we-get-here), but I'm sure there could be a better blog post about the ESM/CJS situation ? I'm honestly not confident enough to write a good one

@TheAlexLichter 🙌

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,10 +8,10 @@ import {
   type OutputOptions,
   type RolldownPluginOption,
 } from 'rolldown'
-import satisfies from 'semver/functions/satisfies.js'
 import { exec } from 'tinyexec'
 import treeKill from 'tree-kill'
 import { attw } from './features/attw'
+import { warnLegacyCJS } from './features/cjs'
 import { cleanOutDir } from './features/clean'
 import { copy } from './features/copy'
 import { writeExports, type TsdownChunks } from './features/exports'
@@ -48,23 +48,6 @@ import type { Options as DtsOptions } from 'rolldown-plugin-dts'
  */
 export async function build(userOptions: Options = {}): Promise<void> {
   const { configs, files: configFiles } = await resolveOptions(userOptions)
-
-  // If one of the config includes the `cjs` format and one of its target is higher than node 23.0.0/22.12.0, warn the user about the deprecation of CommonJS.
-  const config = configs.find((config) =>
-    config.format.includes('cjs') && config.target
-      ? config.target.some((t) =>
-          satisfies(t.split('node')[1], '>=23.0.0 || >=22.12.0'),
-        )
-      : false,
-  )
-  if (config) {
-    // Use the logger associated with the config
-    config.logger.warn(
-      'We recommend using the ESM format instead of CommonJS.\n' +
-        'ESM format is compatible with every platform and runtime, and most libraries now ship only ESM modules.\n' +
-        'See more at https://nodejs.org/en/learn/modules/publishing-a-package#how-did-we-get-here',
-    )
-  }
 
   let cleanPromise: Promise<void> | undefined
   const clean = () => {
@@ -120,6 +103,8 @@ export async function buildSingle(
   let ab: AbortController | undefined
 
   const { hooks, context } = await createHooks(config)
+
+  warnLegacyCJS(config)
 
   await rebuild(true)
   if (watch) {
--- a/src/features/cjs.ts
+++ b/src/features/cjs.ts
@@ -1,0 +1,28 @@
+import coerce from 'semver/functions/coerce.js'
+import satisfies from 'semver/functions/satisfies.js'
+import type { ResolvedOptions } from '../options'
+
+/**
+ * If the config includes the `cjs` format and
+ * one of its target >= node 23.0.0 / 22.12.0,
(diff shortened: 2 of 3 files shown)


src/index.ts
```diff
@@ -47,6 +47,17 @@ export async function build(userOptions: Options = {}): Promise<void> {
 
   const { configs, files: configFiles } = await resolveOptions(userOptions)
 
+  // If one of the config includes the `cjs` format, warn the user about its deprecation.
+  if (configs.some((config) => config.format.includes('cjs'))) {
+    logger.warn(
+      'We recommend using the `es` format instead of `cjs`. ' +
```
sxzz:
```suggestion
      'We recommend using the ESM format instead of CommonJS. ' +
```

and below

src/index.ts
```diff
@@ -47,6 +47,17 @@ export async function build(userOptions: Options = {}): Promise<void> {
...
   const { configs, files: configFiles } = await resolveOptions(userOptions)
 
+  // If one of the config includes the `cjs` format, warn the user about its deprecation.
+  if (configs.some((config) => config.format.includes('cjs'))) {
+    logger.warn(
+      'We recommend using the ESM format instead of CommonJS. ' +
+        '\n' +
+        'ESM format is compatible with every platform and runtime, and most libraries now ship only ESM modules.' +
+        '\n' +
+        'See more at https://nodejs.org/fr/learn/modules/publishing-a-package#how-did-we-get-here',
```
sxzz: Use English by default.
```suggestion
      'We recommend using the ESM format instead of CommonJS.\n' +
        'ESM format is compatible with every platform and runtime, and most libraries now ship only ESM modules.\n' +
        'See more at https://nodejs.org/en/learn/modules/publishing-a-package#how-did-we-get-here',
```

Gugustinette:
Not sure about my target/semver usage @sxzz 😭

src/index.ts
```diff
@@ -43,6 +44,23 @@ import type { Options as DtsOptions } from 'rolldown-plugin-dts'
...
+  if (
+    configs.some((config) =>
+      config.format.includes('cjs') && config.target
+        ? config.target.some((t) =>
+            satisfies(t.split('node')[1], '>=23.0.0 || >=22.12.0'),
+          )
+        : false,
+    )
+  ) {
+    logger.warn(
```
TheAlexLichter: Glad to see this PR!

One thing that I'd suggest is having an option/env var/anything to disable the warning and telling people about it. We can always "enforce" it further with time passing (no way to disable the warning, etc. etc.).

I'm not entirely sure if that is necessary or if people are bothered by the warning, I can only imagine that some might be (and migrations away from CJS often take time due to new majors, etc. etc.).

Open for thoughts/feedback though!

Gugustinette: Yeah I thought about it, not sure how to tackle this.

I considered the `silent` option a decent workaround, but at this point it might be worth introducing a more flexible logger ? Being able to failOnWarn or hide only warnings,...

sxzz: Good point! Maybe we should have `logLevel` and `logger` (custom logger) options first. 

Also warn only once, if there are multiple configs

Gugustinette: We could wrap everything in a dedicated option ?

```ts
{
  logger: {
    silent: boolean
    logLevel: 'info' | 'warning' | 'error'
    handler: (...args) => {}
  }
}
```

Should probably be a dedicated issue tbh 😭

sxzz: Yes, a dedicated issue
- Keep them at the top-level
- Add `silent` to `logLevel` (Vite), and deprecate `silent` option (override logLevel for now).
- Rename `handler` to `customLogger` (Vite)
    ```ts
    interface Logger {
	  info(msg: string, options?: LogOptions): void;
	  warn(msg: string, options?: LogOptions): void;
	  warnOnce(msg: string, options?: LogOptions): void;
	  error(msg: string, options?: LogErrorOptions): void;
	  clearScreen(type: LogType): void;
	  hasErrorLogged(error: Error | RollupError): boolean;
	  hasWarned: boolean;
    }
    ```

If there are multiple configs with different logger option, it's a very complicated case.

sxzz: https://github.com/rolldown/tsdown/issues/405

PR #404, opened 2025-08-01
feat: support silent as programmatic usage

### LINKED ISSUE #359: silent flag is only respected for programmatic usage and not cli usage
## Broken:

Any `tsdown.config.ts` file:

```ts
import { defineConfig } from 'tsdown';
export default defineConfig({
   ...
   silent: true
});
```

CLI: `tsdown`

Result: silent flag is **not** respected

___

## Working:

Any `tsdown.ts` file:

```ts
import { build } from 'tsdown';
await build({
   ...
   silent: true
});
```

CLI: `tsx tsdown.ts`

Result: silent flag is respected

### What is expected?

`silent: true` should be respected for both CLI usage and programmatic usage

`silent: true` is not respected for CLI usage of `tsdown`

### Any additional comments?

_No response_

Gugustinette:
### Description

Wasn't sure how to correctly implement this one, it pretty much implies to load the config file before logging anything, that makes logging a bit annoying at the beginning of the process.

### Linked Issues

Resolves #359

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,6 +1,8 @@
 import process from 'node:process'
+import { dim } from 'ansis'
 import { cac } from 'cac'
 import debug from 'debug'
+import { VERSION as rolldownVersion } from 'rolldown'
 import { version } from '../package.json'
 import { resolveComma, toArray } from './utils/general'
 import { logger } from './utils/logger'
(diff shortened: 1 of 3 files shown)


sxzz (commented):
I made some tweaks:

- The tsdown and rolldown versions are only displayed in the CLI, and they can only be controlled using a CLI flag.  
- Even if silent mode is enabled via file configuration, the versions will always be displayed and cannot be disabled through file configuration.  
- Waiting until the configurations are resolved seems very slow, and that's also how Vite operates.

sxzz (approved):

PR #411, opened 2025-08-04
feat: add failOnWarn option

### LINKED ISSUE #322: Option to treat warnings as error

If I do a tsdown build, I can get
```
[UNRESOLVED_IMPORT] Warning: Could not resolve 'your-library/some-export' in src/index.ts
```
and this is treated as an external dependency.

This isn't something I want, I'd rather fail!

### Suggested solution

--warnings-as-errors? --fail-on-resolve-errors? Some way to fail instead of printing warnings and exiting with 0.

### Alternative

_No response_

### Additional context

_No response_

### Validations

Gugustinette:
### Description

Added a `failOnWarn` option by making the logger throw on `warn()` and `warnOnce()`.

Not sure if the logger being responsible for throwing is the best way to handle this however 😕

### Linked Issues

Resolves #322

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -61,11 +61,10 @@ export function createLogger(
     },
 
     warn(...msgs: any[]): void {
-      if (failOnWarn) {
-        throw new Error(msgs.join(' '))
-      }
-
       const message = format(msgs)
+      if (failOnWarn) {
+        throw new Error(message)
+      }
       warnedMessages.add(message)
       output('warn', `\n${bgYellow` WARN `} ${message}\n`)
     },
@@ -77,7 +76,7 @@ export function createLogger(
       }
 
       if (failOnWarn) {
-        throw new Error(msgs.join(' '))
+        throw new Error(message)
       }
       warnedMessages.add(message)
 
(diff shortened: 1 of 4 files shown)


src/utils/logger.ts
```diff
@@ -34,6 +34,7 @@ const warnedMessages = new Set<string>()
 
 export function createLogger(
   level: LogLevel = 'info',
+  failOnWarn: boolean = false,
   { customLogger, console = globalThis.console }: LoggerOptions = {},
```
sxzz: Move `failOnWarn` to `LoggerOptions`

src/utils/logger.ts
```diff
@@ -56,6 +61,10 @@ export function createLogger(
     },
 
     warn(...msgs: any[]): void {
+      if (failOnWarn) {
+        throw new Error(msgs.join(' '))
```
sxzz:
```suggestion
        throw new Error(message)
```

src/utils/logger.ts
```diff
@@ -66,6 +75,10 @@ export function createLogger(
       if (warnedMessages.has(message)) {
         return
       }
+
+      if (failOnWarn) {
+        throw new Error(msgs.join(' '))
```
sxzz: Same here

sxzz (changes requested):

sxzz (approved):

Issue #418, opened 2025-08-07
When both the `--dts` CLI argument and the `dts` option in the configuration file are specified, the `dts` option will be overridden


hotdogc1017:
### Reproduction link or steps

repo: https://github.com/hotdogc1017/tsdown-bug-report

### What is expected?

When specified simultaneously, the configuration should be merged. Because errors may occur in some edge cases.

### What is actually happening?

The [bug report](https://github.com/hotdogc1017/tsdown-bug-report) repository is a Vue project, so I manually configured the dts option in `tsdown.config.ts`. Here is the complete configuration file:

```ts
import { defineConfig } from "tsdown";
import vue from "unplugin-vue/rolldown";

export default defineConfig({
  plugins: [vue()],
  dts: {
    vue: true,
  },
});
```

I have prepared two scripts:

  - `pnpm build`: `tsdown build`
  - `pnpm build:dts`: `tsdown build --dts`

The only difference between them is the `--dts` option.

Running `pnpm build` works fine, but running `pnpm build:dts` results in the following error:

```text
[plugin rolldown-plugin-dts:generate]
Error: Unable to load file /home/hotdog/bug-report/tsdown-bug-report/src/app.vue from the program. This seems like a bug of rolldown-plugin-dts. Please report this issue to https://github.com/sxzz/rolldown-plugin-dts/issues
```

When I checked the source code, I found that the `--dts` option overrides the `dts` value in the configuration file. For this reason, I have already prepared a PR to fix it.

sxzz:
Use `--dts.vue` instead.

PR #419, opened 2025-08-07
fix: The `dts` value in the configuration file is prioritized over the one specified on the command line.

hotdogc1017:
fix: #418


sxzz:
It will break `--no-dts` cli flag.

PR #438, opened 2025-08-15
feat: detect indentation when migrating package.json

### LINKED ISSUE #437: package.json spacing rewritten after `migrate`

- Open a `tsup` project with a `package.json` formatted with hard tabs
- Run `npx tsdown migrate`

### What is expected?

`tsup` is removed, and `tsdown` is added, with no changes to whitespace/formatting

Hard tabs are removed from package.json and replaced with spaces for indentation.

### Any additional comments?

_No response_

Cherry:
### Description

When performing a migration of `package.json` during `npx tsdown migrate`, previously the `package.json` would always output with 2 spaces for indentation. This PR (roughly) detects the indentation of the `package.json` and tries to maintain it, so that folks who use hard tabs, 4 spaces, etc. can maintain that after migration.

### Linked Issues
Fixes #437 

I included a few tests for this functionality, but couldn't see any other migrate-specific tests. If there's a better spot for these, just let me know.


sxzz (commented):
Can we simply detect the `'\t'` character? In `package.json`, tab characters within values are extremely uncommon.

Cherry:
> Can we simply detect the `'\t'` character? In `package.json`, tab characters within values are extremely uncommon.

Happy to update for that if preferred! I was trying to maintain any number of spaces too - like 2, 3, 4, etc. but if you think it's best to simply detect tab and use `\t` or fallback to 2 spaces, I'm happy to update.

sxzz (approved):

Issue #421, opened 2025-08-09
The CJS information groups printed by the terminal are not combined into one statistic.


btea:
### Reproduction link or steps

`tsdonw.config.ts` With the configuration in place, run the tsdown command directly.
`
export default defineConfig({
  entry: ["src/index.ts"],
  outDir: "dist",
  format: ["cjs", "esm"],
  dts: true,
  clean: true,
  sourcemap: true
})
`

### What is expected?

Hope that the report information about CJS files will be printed together with the statistics.

### What is actually happening?

The printed CJS information is divided into two groups of printing statistics.

<img width="984" height="448" alt="Image" src="https://github.com/user-attachments/assets/43cbc1cb-5d53-4e91-8c1d-17f78a7f01eb" />

sxzz:
Because the DTS for CJS are generated in a separate build process, independent of the CJS bundling itself, the order in which these two processes finish cannot be guaranteed. I believe the current output should be sufficient, especially since most mainstream libraries are now migrating to ESM-only.

PR #441, opened 2025-08-17
fix: combine cjs information groups

ronantakizawa:
### Description

  This PR fixes the issue where CJS files were being reported in separate groups when building with both CJS
  format and DTS enabled. Previously, the build output would show CJS files split across multiple reporting
  sections, making it difficult to understand the total size and file count for the CJS build.

  **Changes made:**

  1. **Modified `getBuildOptions` in `src/features/rolldown.ts`** to accept a `skipReport` parameter that
  prevents the ReportPlugin from being added to the initial CJS build when DTS is enabled.
  2. **Updated build logic in `src/index.ts`** to skip reporting for the first CJS build when DTS is enabled, and
   instead manually generate a combined report after both builds complete.
  3. **Extracted reporting logic** from ReportPlugin into a standalone `generateReport` function in
  `src/features/report.ts` to enable manual reporting with combined file data.
  4. **Added conditional logic** to ensure the fix only affects CJS builds with DTS enabled, leaving all other
  build scenarios unchanged

### Linked Issues
  Fixes issue #421

### DIFF
--- a/src/features/report.ts
+++ b/src/features/report.ts
@@ -8,7 +8,7 @@ import { RE_DTS } from 'rolldown-plugin-dts/filename'
 import { formatBytes } from '../utils/format'
 import { noop } from '../utils/general'
 import { prettyFormat, prettyName, type Logger } from '../utils/logger'
-import type { OutputAsset, OutputChunk, Plugin } from 'rolldown'
+import type { OutputAsset, OutputBundle, OutputChunk, Plugin } from 'rolldown'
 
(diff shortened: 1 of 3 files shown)


sxzz:
Because the DTS for CJS are generated in a separate build process, independent of the CJS bundling itself, the order in which these two processes finish cannot be guaranteed. I believe the current output should be sufficient, especially since most mainstream libraries are now migrating to ESM-only.

PR #517, opened 2025-09-28
feat!: defaults `fixedExtension` to true when platform is `node`

sxzz:
### Description

Defaults `fixedExtension` to true when `platform` is `node`.

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -210,6 +210,7 @@ async function resolveUserConfig(
     cjsDefault = true,
     globImport = true,
     inlineOnly,
+    fixedExtension = platform === 'node',
   } = userConfig
 
   const logger = createLogger(logLevel, { customLogger, failOnWarn })
@@ -343,6 +344,7 @@ async function resolveUserConfig(
     cjsDefault,
     globImport,
     inlineOnly,
+    fixedExtension,
   }
 
   return config
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -287,7 +287,8 @@ export interface UserConfig {
    * Use a fixed extension for output files.
    * The extension will always be `.cjs` or `.mjs`.
    * Otherwise, it will depend on the package type.
-   * @default false
+   *
+   * Defaults to `true` if `platform` is set to `node`, `false` otherwise.
    */
   fixedExtension?: boolean
 
@@ -504,7 +505,6 @@ export type ResolvedConfig = Overwrite<
     | 'alias'
     | 'external'
     | 'onSuccess'
-    | 'fixedExtension'
     | 'outExtensions'
     | 'hooks'
     | 'copy'
--- a/tests/__snapshots__/banner-and-footer-option.snap.md
+++ b/tests/__snapshots__/banner-and-footer-option.snap.md
@@ -1,14 +1,14 @@
-## index.d.ts
+## index.d.mts
 
-```ts
+```mts
 // dts banner
 export { };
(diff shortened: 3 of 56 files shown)


RobinTail:
This should depend on `type` in `package.json` by default

sxzz:
See also https://bsky.app/profile/sxzz.dev/post/3lzwm2sgaus2z

silverwind:
I see `.mjs` as a legacy solution and strongly prefer `.js`, so this change caught me by surprise.

It broke my lib because in package.json I have `"exports": "./dist/index.js"` which after this change pointed to a non-existing file.

kitschpatrol:
I personally agree with [Bjorn Lu's take](https://bsky.app/profile/bluwy.me/post/3lzwqab7wn22y) on this — but if this is to be the default behavior in tsdown, maybe the `fixedExtension` option flag can be exposed in the CLI?

I had a bunch of projects that worked perfectly with a vanilla `tsdown` command and no additional configuration, which will now require a `tsdown.config.ts` file to bypass this default. Appending a flag would be cleaner for these cases.

Thanks for your work!

sxzz:
- All options can be overridden via CLI flags. To disable a boolean option, use `--no-fixed-extension` (`--no-<option-name>`).
- You can also enable `exports` option to automatically generate `exports` field on package.json.

Thank you for your understanding.

kitschpatrol:
Thanks! I didn't realize all the options are available as CLI flags, thought it was limited to https://tsdown.dev/reference/cli and what's documented in the `--help` output.

molvqingtai:
Although configuration options are provided, it is still hoped that the default should follow the package.json type, or it should be the default --no-fixed-extension, generating code that is more in line with expectations in the case of minimal configuration or no configuration.

sxzz:
v0.16 has been released for some time now, so we can no longer change the default behavior without affecting those who have already migrated. In the future, tsdown will be more cautious when introducing breaking changes.

NiklasDah:
> v0.16 has been released for some time now, so we can no longer change the default behavior without affecting those who have already migrated. In the future, tsdown will be more cautious when introducing breaking changes. (@sxzz)

Would it at least be possible to update the breaking-changes section to clarify the impact of this change? The current entry:
> Defaults `fixedExtension` to `true` when platform is `node`

doesn’t explain **why** this is breaking. I overlooked it and ran into unexpected bugs even after checking the release notes. Adding something like
> this means that no .js-files will be created by default

retroacively, would help users better understand the impact when skimming the changelog.

sxzz:
The changelog has been updated in the release notes: https://github.com/rolldown/tsdown/releases/tag/v0.16.0

Thank you for your feedback!

DopamineDriven:
This should definitely depend on type in package.json by default. That's by convention

this breaks pipelines unnecessarily

silverwind:
I'd make it by default match the extension of the entry file. And maybe add a option `outputExtension` or similar so users can specify the extension they want.

sxzz:
You can use `outExtensions`

kh4f:
Hi @sxzz. Would you consider switching the default artifact extension back to `.js` in the upcoming v0.23? It just looks a bit cleaner, and ESM is pretty much the default choice today.

sxzz:
No, see https://x.com/_pi0_/status/2093283873165545832

kh4f:
> No, see https://x.com/_pi0_/status/2093283873165545832

Didn't know that, thanks

PR #535, opened 2025-10-13
feat(create-tsdown): migrate from Gugustinette/create-tsdown

sxzz:
Migrating from https://github.com/Gugustinette/create-tsdown, thanks to @Gugustinette !!!!


Gugustinette:
@sxzz I've changed `create-tsdown` npm trusted publisher settings to the `rolldown/tsdown` repo 👍

PR #559, opened 2025-10-24
fix(exports): correctly generate export paths for nested index files

### LINKED ISSUE #558: export true doesn't normalize subpackages exports paths on Windows

Add a subpackage export on Windows laptop enabling `export: true` in the `tsdown.config.ts` file:

screenshot using \ in the subpackage exports

debug generateExports screenshot

 

### What is expected?

Subpackage exports should use `/` instead separator path

Using separator path, requires applying a normalization

### Any additional comments?

I'm going to send a PR to fix this bug.

In fact the generated subpackages exports are wrong: we need to handle multiple entries and group them: check core/types and core/index, maybe a side effect of this issue (?)

userquin:
### Description

This fix normalizes entrypoint paths to correctly handle subpackage exports, especially for nested index files and paths generated on Windows.

### Linked Issues

resolves #558

### Additional context

~~After updating my fork, on my local there are 3/4 tests failing: env and shims related to define/inject changes at Rolldown~~

### DIFF
--- a/src/features/exports.ts
+++ b/src/features/exports.ts
@@ -111,13 +111,13 @@ export async function generateExports(
         (chunk) =>
           chunk.type === 'chunk' &&
           chunk.isEntry &&
-          !RE_DTS.test(slash(chunk.fileName)),
+          !RE_DTS.test(chunk.fileName),
       ).length === 1
     for (const chunk of chunksByFormat) {
       if (chunk.type !== 'chunk' || !chunk.isEntry) continue
 
       const normalizedName = slash(chunk.fileName)
-      const ext = path.extname(normalizedName)
+      const ext = path.extname(chunk.fileName)
       let name = normalizedName.slice(0, -ext.length)
 
       const isDts = name.endsWith('.d')


src/features/exports.ts
```diff
@@ -111,20 +111,21 @@ export async function generateExports(
         (chunk) =>
           chunk.type === 'chunk' &&
           chunk.isEntry &&
-          !RE_DTS.test(chunk.fileName),
+          !RE_DTS.test(slash(chunk.fileName)),
```
sxzz: This line can be reverted, since suffix name is the same in both Windows and Linux.

src/features/exports.ts
```diff
@@ -111,20 +111,21 @@ export async function generateExports(
...
       ).length === 1
     for (const chunk of chunksByFormat) {
       if (chunk.type !== 'chunk' || !chunk.isEntry) continue
 
-      const ext = path.extname(chunk.fileName)
-      let name = chunk.fileName.slice(0, -ext.length)
+      const normalizedName = slash(chunk.fileName)
+      const ext = path.extname(normalizedName)
+      let name = normalizedName.slice(0, -ext.length)
```
sxzz, line 121: Same here?

src/features/exports.ts
```diff
@@ -111,20 +111,21 @@ export async function generateExports(
...
 
       const isDts = name.endsWith('.d')
       if (isDts) {
         name = name.slice(0, -2)
       }
       const isIndex = onlyOneEntry || name === 'index'
-      const distFile = `${outDirRelative ? `./${outDirRelative}` : '.'}/${chunk.fileName}`
+      const distFile = `${outDirRelative ? `./${outDirRelative}` : '.'}/${normalizedName}`
```
sxzz, line 128: Seems like only here needs be normalized.

userquin: Will apply changes tmr

src/features/exports.ts
userquin: uhmmm, at L120, using chunk.fileName at L121 error:

error applying suggestion

PR #561, opened 2025-10-25
refactor: consolidate template options to have single source of truth

yangshun:
### Description

The template options were being repeated in multiple places, and one part has gone out of sync (the CLI `template` argument was missing the `minimal` from the available templates). 

```
Error: Invalid template "foo". Available templates: default, vue, react, solid
```

Consolidate the template options into a single array and use it everywhere it's needed.

### Test Plan

Built the CLI, tested the script. The available options and error messages show up fine.

image

### DIFF
--- a/packages/create-tsdown/src/index.ts
+++ b/packages/create-tsdown/src/index.ts
@@ -93,7 +93,7 @@ export async function resolveOptions(
   } else {
     template = await select({
       message: 'Which template do you want to use?',
-      options: templateOptions.slice(),
+      options: [...templateOptions],
       initialValue: 'default',
     })
 


packages/create-tsdown/src/index.ts
```diff
@@ -74,21 +84,16 @@ export async function resolveOptions(
 
   let template: Options['template'] | symbol = options.template
   if (template) {
-    if (!['default', 'minimal', 'vue', 'react', 'solid'].includes(template)) {
+    const templateOptionsValues = templateOptions.map((option) => option.value)
+    if (!templateOptionsValues.includes(template)) {
       throw new Error(
-        `Invalid template "${template}". Available templates: default, vue, react, solid`,
```
yangshun, line 79: Missing the `minimal` template

packages/create-tsdown/src/index.ts
```diff
@@ -74,21 +84,16 @@ export async function resolveOptions(
...
     template = await select({
       message: 'Which template do you want to use?',
-      options: [
-        { value: 'default', label: 'Default' },
-        { value: 'minimal', label: 'Minimal' },
-        { value: 'vue', label: 'Vue' },
-        { value: 'react', label: 'React' },
-        { value: 'solid', label: 'Solid' },
-      ],
+      options: templateOptions.slice(),
```
yangshun: Had to make a copy as `select()` only accepts mutable arrays

sxzz (approved):

Issue #247, opened 2025-05-20
Docs Roadmap


sxzz:
- Programmatic Usage
- Workspace
- Benchmark https://github.com/rolldown/tsdown/issues/94
- Starter Templates https://github.com/rolldown/tsdown/issues/128

mamlzy:
monorepo best practice, please😉

PR #569, opened 2025-10-28
docs: Mention typescript feature support / decorators

janis-me:
### Description

Update the `target` docs page to mention that it's required to change the `tsconfig` settings for certain typescript features, like TC39 decorators, as requested in https://github.com/rolldown/tsdown/issues/247

I added a new section `Typescript feature compatibility` on the `options/target` page, because it's somewhat related and I didn't want to create a new page for this. Please let me know if I should change the naming, document more features etc.

### Linked Issues

- Issue mentioning that it's needed to change tsconfig: https://github.com/rolldown/tsdown/discussions/271
- Docs TODO issue: https://github.com/rolldown/tsdown/issues/247

### Additional context

I tested that the links work, and added the link to the `tsconfig` tsdown config option as an external link

### DIFF
--- a/docs/options/target.md
+++ b/docs/options/target.md
@@ -77,6 +77,12 @@ You can also pass an array of targets to ensure compatibility across multiple en
 tsdown --target chrome100 --target node20.18
 ```
 
+# Typescript feature compatibility
+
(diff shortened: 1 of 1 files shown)


sxzz:
Actually, tsdown may not respect certain tsconfig fields, such as `target` and `module`. Therefore, this documentation could potentially mislead users.

janis-me:
Makes sense, but you still want to let users know how to enable, for example, decorator support, right? (As mentioned in #247 ) Any idea on where to put that docs?

janis-me:
Ah, I see the docs have already been updated. Thanks!

PR #581, opened 2025-11-04
feat(entry): infer file extensions for entry paths

jinghaihan:
### Description

I’ve been working on migrating from **unbuild** to **tsdown** recently, and I noticed that the `entries` option in **unbuild** can omit file extensions.
Therefore, I added the same inference feature to the `entry` option in **tsdown**.  
The inference will follow the order below:

```
const DEFAULT_EXTENSIONS: string[] = [
  '.ts',
  '.tsx',
  '.mts',
  '.cts',
  '.mjs',
  '.cjs',
  '.js',
  '.jsx',
  '.json',
]
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,18 +2,17 @@ node_modules
 dist
 .temp
 temp
+.idea
+.netlify
+.rolldown
 
 *.log
 .DS_Store
 .eslintcache
 tsdown-*.tgz
-.idea
 
 docs/.vitepress/dist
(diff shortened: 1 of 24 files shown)


sxzz:
Thanks for your contribution! I implemented this feature with a simpler way, and used your test code. See https://github.com/rolldown/tsdown/commit/2c525d440a11c0b0dd7d423b1205dcad9c905310

PR #585, opened 2025-11-07
perf: switch `unconfig-core` & mark `unrun` stable

sxzz:

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -148,10 +148,8 @@ function resolveConfigLoader(
       process.versions.deno
     )
     return nativeTS ? 'native' : 'unrun'
-  } else if (configLoader === 'native') {
-    return 'native'
   } else {
-    return 'unrun'
+    return configLoader === 'native' ? 'native' : 'unrun'
   }
 }
 


docs/options/config-file.md
```diff
@@ -66,13 +66,12 @@ This is useful if you want to rely solely on command-line options or default set
 
 `tsdown` supports multiple config loaders to accommodate various file formats. You can select a config loader using the `--config-loader` option. The available loaders are:
 
-- `auto` (default): Utilizes native runtime loading for TypeScript if supported; otherwise, defaults to `unconfig`.
+- `auto` (default): Utilizes native runtime loading for TypeScript if supported; otherwise, defaults to `unrun`.
 - `native`: Loads TypeScript configuration files using native runtime support. Requires a compatible environment, such as the latest Node.js, Deno, or Bun.
-- `unconfig`: Loads configuration files with the `unconfig` library, ensuring broad compatibility across different runtimes.
-- `unrun` _(experimental)_: Loads configuration files using the [`unrun`](https://gugustinette.github.io/unrun/) library, which provides similar compatibility as `unconfig` but with more performances. Install the `unrun` package to use this loader.
+- `unrun`: Loads configuration files using the [`unrun`](https://gugustinette.github.io/unrun/) library. Install the `unrun` package to use this loader.
```
RobinTail, line 71: > Install the `unrun` package to use this loader.

it's now installed by default as direct runtime dependency.
which reverts the fix made in #570 , @sxzz

sxzz: Yes. I will update the docs.

sxzz: tsdown and unrun will ensure that the rolldown versions remain consistent before release, up until rolldown officially launches its stable version.

RobinTail: I understand, my point is that if I'm using `.ts` extensions in imports so that don't even need `unrun`, it's still installed anyway @sxzz

sxzz: Here are the reasons why we currently keep `unrun`:

- Node.js 20 does not support importing TypeScript files directly, and it's unreasonable to require users to install additional tools manually.
- Importing TypeScript files requires an explicit `.ts` file extension if using native loader, and importing JSON files needs `with { type: 'json' }`, both of which are incompatible with the current Vite/Vitest config resolver.
- `unrun` is lightweight, and both tsdown and unrun rely on the Rolldown engine.
- Previously, unconfig used jiti by default. Since Rolldown is already used in tsdown, its bundle size is larger than unrun.

RobinTail: ok, thank you for the clarification

Issue #634, opened 2025-12-08
Support glob in `copy`


kricsleo:
### Clear and concise description of the problem

Often, the resources we need to copy are not centralized in a specific folder, but are instead placed alongside the source code.
A common scenario in a UI library involves placing `./button.css` and `./button.ts` files side-by-side. In `unbundle: true` mode, `ts` files can maintain their directory structure within the `dist` folder, but CSS files (and other resources) also need to be copied to their corresponding locations in the output directory. 
By supporting glob patterns in `copy`, we can easily achieve this.
Since we already support the `copy` feature, also supporting `glob` patterns would be a bonus, saving developers from installing another copy plugin like [rollup-plugin-copy](https://github.com/vladshcherbin/rollup-plugin-copy).

For example (https://github.com/rolldown/tsdown/issues/160#issuecomment-2848230653):

```ts
export default defineConfig({
  copy: [
    'src/**/*.css',
    { from: 'src/**/*.css', to: 'dist', flatten: false }
  ],
});
```

### Suggested solution

We could support this by globbing all matched files (utilizing `tinyglobby`), copying them to the destination folder, and offering a `flatten` option to indicate whether they should be flattened or retain their original folder structure.

### Additional context

I'd like to submit a PR for this if that's what we'd like to have ❤️

### Validations

- [x] Follow our [Code of Conduct](https://github.com/sxzz/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/sxzz/contribute).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sxzz:
PR welcome if anyone wants to work on that!

PR #637, opened 2025-12-10
feat(copy): support glob in `copy`

kricsleo:
### Description

This PR enables glob support in `copy` feature.

### Linked Issues

resolves https://github.com/rolldown/tsdown/issues/634

```ts
export default defineConfig({
  copy: [
    'src/assets', // src/assets -> dist/assets
    'src/assets/foo.txt', // src/assets/foo.txt -> dist/assets/foo.txt
    'src/styles/**/*.css', // src/styles/foo/bar.css -> dist/styles/foo/bar.css
    { from: 'src/styles/**/*.css', to: 'dist/qux' }, // src/styles/foo/bar.css -> dist/qux/styles/foo/bar.css 
    { from: 'src/styles/**/*.css', to: 'dist/qux', flatten: true }, // src/styles/foo/bar.css -> dist/qux/bar.css
  ]
})
```

### DIFF
--- a/src/features/copy.ts
+++ b/src/features/copy.ts
@@ -2,7 +2,6 @@ import path from 'node:path'
 import { glob, isDynamicPattern } from 'tinyglobby'
 import { fsCopy, fsStat } from '../utils/fs.ts'
 import { toArray } from '../utils/general.ts'
-import { prettyName } from '../utils/logger.ts'
 import type { ResolvedConfig } from '../config/index.ts'
 import type { Arrayable, Awaitable } from '../utils/types.ts'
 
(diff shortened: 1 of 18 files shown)


src/features/copy.ts
```diff
@@ -43,4 +69,28 @@ export async function copy(options: ResolvedConfig): Promise<void> {
...
+      const isFile = (await fsStat(from))?.isFile()
+      const to = isFile ? path.join(dest, parsedFrom.base) : dest
+      return [from, to]
+    }
+
+    const to = path.join(
+      dest,
+      // Stripe off the first segment to avoid unnecessary nesting
+      // e.g. "src/index.css" -> index.css" -> "dist/index.css"
+      parsedFrom.dir.replace(parsedFrom.dir.split(path.sep)[0], ''),
```
fengmk2, line 90: There should be an breaking change here.

The previous configuration
```ts
copy: [
    {
      from: 'src/lib/onerror_page.mustache.html',
      to: 'dist/lib/onerror_page.mustache.html',
    },
  ],
```

After the change, it can only be rewritten as
```ts
copy: [
    {
      from: 'src/lib/onerror_page.mustache.html',
      to: 'dist/lib',
    },
  ],
```

sxzz: Yes, it's a breaking change. See https://github.com/rolldown/tsdown/issues/650#issuecomment-3654281292

PR #654, opened 2025-12-17
feat: add CSS code splitting support

jinghaihan:
### Description

This PR adds CSS code splitting support to `tsdown`, allowing users to merge all CSS files into a single `style.css` file when `cssCodeSplit: false` is enabled.

**Motivation:**

Bundlers typically don't automatically import CSS files, which creates complexity for library developers using Vue SFCs when building async components. After reviewing Vite's implementation, I ported their `cssCodeSplit` option to `tsdown` to simplify CSS handling for library consumers.

**Implementation:**

- Added `cssCodeSplit` configuration option (defaults to `true` to preserve existing behavior)
- CSS is merged in the order: synchronous imports first (lowest priority), then dynamic imports (highest priority)
- Individual CSS files are removed from the bundle after merging

### Additional context

**Testing scope:**
I primarily work with Vue development, so this feature has been tested in Vue projects.

**Default value consideration:**
Given that `tsdown` is primarily used as a library bundler, there might be a case for making `cssCodeSplit: false` the default value, as it simplifies CSS handling for library consumers.

**Credits:**
This implementation is based on Vite's CSS code splitting feature.


docs/options/css.md
```diff
@@ -0,0 +1,37 @@
+# CSS
+
+Configure CSS generation options for your build output.
+
+## Options
```
sxzz: Options are generated in API reference. Please remove them.

src/config/types.ts
```diff
@@ -466,6 +467,11 @@ export interface UserConfig {
    */
   exports?: WithEnabled<ExportsOptions>
 
+  /**
+   * Configure CSS generation options.
+   */
+  css?: WithEnabled<CssOptions>
```
sxzz: css cannot be false or disabled. It's meaning less.

sxzz:
```suggestion
  css?: CssOptions
```

src/config/types.ts
```diff
@@ -563,6 +569,7 @@ export type ResolvedConfig = Overwrite<
     ignoreWatch: Array<string | RegExp>
     noExternal?: NoExternalFn
     inlineOnly?: Array<string | RegExp>
+    css: false | CssOptions
```
sxzz:
```suggestion
    css: CssOptions
```

src/features/css.ts
```diff
@@ -0,0 +1,175 @@
...
+ * CSS Code Split Plugin
+ *
+ * When css.splitting is false, this plugin merges all CSS files into a single file.
+ * When css.splitting is true (default), CSS code splitting is preserved.
+ * Based on Vite's implementation.
+ */
+export function CssCodeSplitPlugin(
+  config: Pick<ResolvedConfig, 'css'>,
+): Plugin | undefined {
+  if (config.css === false || config.css?.splitting) return undefined
```
sxzz:
```suggestion
  if (config.css?.splitting) return undefined
```

sxzz: Default value of `config.css?.splitting` is true. I'm not sure if handled properly.

EDM115:
that's so cool !

PR #659, opened 2025-12-19
feat(migrate): add monorepo support for tsup migration script

### LINKED ISSUE #452: Monorepo support for `tsup` migration script

Right now, the migration script works well for single repositories. When running it in a monorepo context, it does not do much though and must be run per package.

### Suggested solution

Make the script monorepo-aware + use `workspace` mode where applicable.

### Alternative

_No response_

### Additional context

_No response_

### Validations

lauigi:
### Description

Add monorepo support for the `tsup` migration script. If executed in the monorepo root, the script will run sequentially in the current directory and in each package directory. In all other cases, it will still run only in the specified directory.

### Linked Issues
Resolves #452

### Additional context

The testing fixtures of this PR might not be in the best shape. Need comment.

### DIFF
--- a/docs/zh-CN/guide/migrate-from-tsup.md
+++ b/docs/zh-CN/guide/migrate-from-tsup.md
@@ -1,26 +1,41 @@
 # 从 tsup 迁移
 
-[tsup](https://tsup.egoist.dev/) 是一个功能强大且广泛使用的打包器，与 `tsdown` 有许多相似之处。`tsup` 基于 [esbuild](https://esbuild.github.io/) 构建，而 `tsdown` 则利用了 [Rolldown](https://rolldown.rs/) 的强大能力，带来更**快速**、更**强大**的打包体验。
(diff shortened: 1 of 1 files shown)


packages/migrate/src/index.ts
```diff
@@ -29,11 +31,53 @@ export async function migrate({ cwd, dryRun }: MigrateOptions): Promise<void> {
 
   if (cwd) process.chdir(cwd)
```
sxzz: Move `process.chdir` inside of the loop

sxzz (commented):
Is `tinyglobby` really necessary? I think shells like bash will also resolve glob patterns.

lauigi:
> Is `tinyglobby` really necessary? I think shells like bash will also resolve glob patterns.

In my local, if I remove tinyglobby and rely directly on shell expansion like this, it would break support for PowerShell on Windows.

```typescript
const baseCwd = process.cwd()
  const cwds =
    dirs && dirs.length > 0
      ? dirs.map((d) => path.resolve(baseCwd, d))
      : [baseCwd]
```

sxzz:
Ah... PowerShell...

sxzz (approved):

PR #664, opened 2025-12-22
feat: add support to load env file

### LINKED ISSUE #652: Add Support for Reading Environment Variables from .env Files

Currently, tsdown allows defining compile-time environment variables through two methods: a configuration file or the `--env.* <value>` command-line option. However, it does not support reading environment variables from .env files yet, which makes it challenging to dynamically manage these constants. Adding this feature would enhance flexibility and streamline the workflow.

I'd be happy to help implement this feature :)

### Suggested solution

1. **Add Command-Line Options**:  
   Introduce the following new command-line options:  
   - `--env-file <file>`: Specifies the path to the .env file to load environment variables from.  
   - `--env-prefix <prefix>`: Defines a prefix for environment variables (default: `TSDOWN`).  

   Additionally, allow the `envPrefix` option to be set in the configuration file.

2. **Integrate dotenv for Environment Variable Resolution**:  
   During the resolution of `UserConfig`, use `dotenv` to load environment variables from the specified .env file. Apply the prefix filter (if provided) for security to ensure only relevant variables are included.

### Alternative

_No response_

### Additional context

_No response_

### Validations

toto6038:
### Description

This PR introduces support for loading environment variables via a file, specifically `--env-file` and `--env-prefix` command line flags and corresponding config options. 

`dotenv` is used to parse env file, and variables are filtered by prefix (default is `TSDOWN_`) to prevent accidental exposure of sensitive information.

Regarding variable priority, explicit CLI arguments take precedence. Specifically, if a user defines a variable using both the `--env.*` flag and within an `--env-file` simultaneously, the value provided via the `--env.*` flag will override the one in the file.

### Linked Issues

Resolves #652 

### Additional context

**Note on Node.js Conflict**: Since Node.js v20.6.0, the runtime includes a native `--env-file` flag. If a user provides an invalid path or forgets to use the `--` separator between Node and tsdown arguments, Node.js may intercept the flag and throw an error before tsdown even executes.


src/config/options.ts
```diff
@@ -164,6 +173,23 @@ export async function resolveUserConfig(
     }
   }
 
+  if (envFile) {
+    // MARK: -  handling envFile
+    const resolvedPath = path.isAbsolute(envFile)
+      ? envFile
+      : path.resolve(cwd, envFile)
```
sxzz: resolve directly

```suggestion
    const resolvedPath = path.resolve(cwd, envFile)
```

src/config/options.ts
```diff
@@ -278,6 +305,25 @@ export async function resolveUserConfig(
   })
 }
 
+function loadEnvFile(filePath: string, envPrefixes: string[], logger: Logger) {
+  const env: Record<string, string> = {}
+  const parsed = parse(fs.readFileSync(filePath))
+
+  if (envPrefixes.includes('')) {
```
sxzz: What's this?

```suggestion
  if (envPrefixes === '') {
```

src/config/types.ts
```diff
@@ -565,6 +576,7 @@ export type ResolvedConfig = Overwrite<
       | 'logLevel' // merge to `logger`
       | 'failOnWarn' // merge to `logger`
       | 'customLogger' // merge to `logger`
+      | 'envFile' // merged to `env`
```
sxzz, line 579: And envPrefix

src/cli.ts
```diff
@@ -117,7 +128,11 @@ cli
   })
 
 export async function runCLI(): Promise<void> {
-  cli.parse(process.argv, { run: false })
+  // prevent CAC from treating '--' as its own option
+  cli.parse(
+    process.argv.filter((arg) => arg !== '--'),
```
sxzz: Revert it

src/config/options.ts
toto6038: This safeguard is implemented to warn users if any of `envPrefixes` is set to empty string using `--env-prefix=""`. Because an empty prefix bypasses the filtering logic, it would cause all variables in the env file to be loaded, which poses risks of exposing sensitive information or secrets.

src/cli.ts
toto6038: The reason for this filter is that when a user calls the CLI via `node -- <path_to_tsdown> [args]`, the `--` separator remains in `process.argv`. CAC treats `--` as a positional argument rather than a separator, which can cause it to miss subsequent flags or fail to match commands. 
Should we handle the Node.js argument separator in a different way to ensure flags are still parsed correctly?

toto6038:
> Use `process.loadEnvFile` instead of `dotenv`.

The main reason for using dotenv over `process.loadEnvFile` is that the built-in Node.js method does not return the parsed results; it directly mutates `process.env`.

This makes it impossible to distinguish between variables specifically defined in the file and those already present in the system environment. For example, if a user runs `TSDOWN_FOO=bar tsdown --env-file .env`, and the .env file contains `TSDOWN_BAR=baz`, `process.loadEnvFile` would mix them both into process.env.

By using dotenv, we get an isolated object of just the file's contents. This allows us to precisely apply prefix filtering and handle precedence logic without accidentally picking up or re-processing system-level variables that shouldn't go through the resolution logic.

src/config/options.ts
toto6038: fixed

src/config/types.ts
toto6038: fixed

sxzz:
@toto6038 I found `util.parseEnv` is also available https://nodejs.org/api/util.html#utilparseenvcontent

src/config/options.ts
```diff
@@ -164,6 +173,21 @@ export async function resolveUserConfig(
     }
   }
 
+  if (envFile) {
+    // MARK: -  handling envFile
```
sxzz: Remove comments

src/config/options.ts
```diff
@@ -1,7 +1,9 @@
+import fs from 'node:fs'
```
sxzz:
```suggestion
import { readFile } from 'node:fs/promises'
```

src/config/options.ts
```diff
@@ -278,6 +302,25 @@ export async function resolveUserConfig(
   })
 }
 
+function loadEnvFile(filePath: string, envPrefixes: string[], logger: Logger) {
```
sxzz:
```suggestion
function loadEnv(filePath: string, envPrefixes: string[], logger: Logger) {
```

sxzz: Also read values from existing `process.env`

src/cli.ts
sxzz: I guess this change is not relate to this PR, right? If there's a bug, please provide a detailed example. (maybe it's a incorrect way to use)

src/config/options.ts
toto6038: If we want to read all variables from `process.env`, we could stick to `process.env.loadEnvFile` as we don't need to distinguish the variables specified in env file anymore.

sxzz: No, `{ ...parsedFromEnvFile (with prefixed), ...currentProcessEnv (with prefixed), ...envsFromTsdownOptions }`

toto6038: I'd like to confirm the intended behavior of the following:

- Is loading process.env with prefix filtering exclusive to when envFile is provided?
- Should the override precedence be `--env` > env file > `process.env`?

Thanks!

sxzz: Intended behaviors:
- Always load `process.env`
- Environment variables from env file and `process.env` should be filtered with prefix.
- precedence
  - `env` option (aka CLI `--env`, optional, highest priority)
  - `process.env` (always load)
  - env file (aka CLI `--env-file`, optional)

src/config/options.ts
toto6038: fixed

src/config/options.ts
toto6038: fixed

toto6038:
I have reworked the merging and filtering logic of environment variables. If things look good, I can proceed to elaborate this in documentation.

sxzz (approved):

Issue #627, opened 2025-12-02
`.css` file with same results in empty `.js` output


silverwind:
### Reproduction link or steps

1. Clone https://github.com/silverwind/tsdown-multi
2. Run `npm i && npx tsdown`

### What is expected?

That both `dist/index.js` and `dist/index.cs` contain the bundled code.

### What is actually happening?

`dist/index.js` is generated with 0 bytes because the entry `index.css` apparently overwrites `index.js`:

```
ℹ tsdown v0.17.0-beta.5 powered by rolldown v1.0.0-beta.52
ℹ config file: /Users/silverwind/git/tsdown-multi/tsdown.config.ts
ℹ entry: index.css
ℹ Build start
ℹ dist/index.js   0.00 kB │ gzip: 0.02 kB
ℹ dist/index.css  0.02 kB │ gzip: 0.04 kB
ℹ 2 files, total: 0.02 kB
✔ Build complete in 9ms
```

### Any additional comments?

The same entry file configuration works with vite and rollup.

silverwind:
https://github.com/rolldown/rolldown/issues/8403 specifically discusses this:

> Goal: When a JS chunk contains only CSS imports (no actual JS exports), remove the empty JS chunk from the bundle and clean up import statements referencing it.

If rolldown removes these empty JS chunks, the issue would be solved. Until then, one workaround would be to add a rolldown plugin inside `tsdown` that does the same.

sxzz:
Please try `npm add https://pkg.pr.new/tsdown@dbe3cc9 -D`
If you have any issues, please provide feedback at https://github.com/rolldown/tsdown/issues/472

silverwind:
Updated in https://github.com/silverwind/tsdown-multi/commit/0c947bee92c1160476e05f4fb5bb3a72cb7b1fd0. It produces a `style.css` now, so no more overwriting which is great:

```
ℹ dist/index.js   0.00 kB │ gzip: 0.02 kB
ℹ dist/style.css  0.02 kB │ gzip: 0.04 kB
```

But I would still expect the output file name to be `index.css`, thought it can be worked around in a post-build `mv` step.

sxzz:
There's an option to change default filename: https://main.tsdown.dev/options/css#custom-file-name

silverwind:
Yeah that works for renaming. I just noticed `index.js` is 0 bytes so the overwriting issues still present with or without that option.

[...]

PR #671, opened 2025-12-26
Fix the issue where CSS entry files generate empty JS files that unexpectedly overwrite other JS outputs with the same base name

liwenka1:
## Summary

Fix the issue where CSS entry files generate empty JS files that unexpectedly overwrite other JS outputs with the same base name.

Closes: https://github.com/rolldown/tsdown/issues/627

## Problem

When using multiple entry files like `index.js` and `index.css`, the CSS entry generates an empty `index.js` file which overwrites the legitimate JS output:

// tsdown.config.js
export default {
  entry: ["./index.js", "./index.css"],
}**Expected output:**
- `dist/index.js` - Contains bundled JS code
- `dist/index.css` - Contains CSS code

**Actual output (before this fix):**
- `dist/index.js` - Empty (0 bytes) ❌
- `dist/index.css` - Contains CSS code

This happens because:
1. Both entries resolve to the same entry name `index` (after stripping extensions)
2. Rolldown generates a JS chunk for every entry, including CSS-only entries
3. The empty JS chunk from the CSS entry overwrites the real JS output

## Root Cause

This is fundamentally a **Rolldown behavior limitation**. Rolldown generates a JS chunk for every entry point, even when the entry is a pure CSS file. Ideally, Rolldown should natively support CSS-only entries without generating empty JS files.

> **Note:** A more elegant long-term solution would be for Rolldown to handle this natively. Consider opening an issue in [rolldown/rolldown](https://github.com/rolldown/rolldown) to request this feature.

## Solution

This PR implements a targeted fix that **only activates when there's a naming conflict**:

### 1. Detect naming conflicts in entry resolution (`entry.ts`)

When multiple entries would resolve to the same base name (e.g., `index.js` and `index.css` both → `index`), CSS entries are given a `.css` suffix to avoid collision:
- `index.js` → entry name: `index`
- `index.css` → entry name: `index.css` (only when conflict exists)

**When there's no conflict, behavior remains unchanged:**
- Single `index.css` entry → entry name: `index` (no change from before)

### 2. Add `CssEntryPlugin` to clean up conflict outputs (`css.ts`)

A new plugin that runs in `generateBundle` phase, **only for conflict cases**:
- **Removes** empty JS files from CSS entries that have `.css` in their name (conflict resolution naming)
- **Renames** CSS files with double extensions (e.g., `index.css.css` → `index.css`)

## Key Design Decisions

- **Minimal impact**: Only affects builds where JS and CSS entries share the same base name
- **Backward compatible**: Single CSS entries (like `entry: 'index.css'`) behave exactly as before
- **All existing tests pass**: No snapshot updates needed

## Test Cases

| Scenario | Behavior |
|----------|----------|
| CSS-only entry (no conflict) | ✅ Unchanged - generates `index.mjs` (empty) + `index.css` |
| JS + CSS same name entries | ✅ Fixed - both `index.js` and `index.css` output correctly |
| Multiple CSS entries | ✅ Unchanged - `foo.mjs`, `bar.mjs`, `foo.css`, `bar.css` |
| JS imports CSS | ✅ Works - CSS extracted, import preserved if needed |

## Changes

- `src/features/entry.ts` - Detect naming conflicts and add `.css` suffix for CSS entries when needed
- `src/features/css.ts` - Add `CssEntryPlugin` to clean up conflict outputs
- `src/features/rolldown.ts` - Register the new plugin

## Related

- Issue: https://github.com/rolldown/tsdown/issues/627
- This is a workaround for Rolldown's behavior. A native fix in Rolldown would be the ideal solution.

### DIFF
--- a/src/features/css.ts
+++ b/src/features/css.ts
@@ -1,4 +1,4 @@
-import { RE_CSS } from 'rolldown-plugin-dts/filename'
+import { RE_CSS, RE_JS } from 'rolldown-plugin-dts/filename'
 import type { ResolvedConfig } from '../config/index.ts'
 import type { OutputAsset, OutputChunk, Plugin } from 'rolldown'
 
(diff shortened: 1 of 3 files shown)


jinghaihan:
I noticed that the empty JS files (e.g., `index.css.js`) are still not being deleted from the output directory.

**Current behavior:**
When building with CSS entry points like `['./src/index.ts', './src/index.css']`, the build generates:

```ts
// index.ts
import './index.css'

export const foo = 1
```

```css
// index.css
.bar {
	a: red;
}
```

- `dist/index.js` ✓
- `dist/index.css.js` (empty file, should be removed) ✗
- `dist/style.css` ✓

**Expected behavior:**
The empty `index.css.js` file should be automatically cleaned up after the build process.

liwenka1:
I submitted a new PR:https://github.com/rolldown/tsdown/pull/674 to try to fix this issue.

PR #680, opened 2025-12-31
feat: generate CSS exports when `css.splitting` is disabled

jinghaihan:
### Description

Automatically generates CSS exports in `package.json` when `css.splitting` is disabled. This enables component libraries to bundle CSS into a single file and expose it via package exports for users to import.

### Additional context

`generateExports` now accepts a single `options` object (containing `exports`, `logger`, `css`, `outDir`) instead of separate `ExportsOptions` and `logger` parameters. This change enables future extensions where exports generation depends on multiple configuration options.

### DIFF
--- a/src/features/pkg/exports.ts
+++ b/src/features/pkg/exports.ts
@@ -158,8 +158,7 @@ export async function generateExports(
         name = name.slice(0, -2)
       }
       const isIndex = onlyOneEntry || name === 'index'
-      const outDirRelative = slash(path.relative(pkgRoot, chunk.outDir))
-      const distFile = `${outDirRelative ? `./${outDirRelative}` : '.'}/${normalizedName}`
+      const distFile = join(pkgRoot, chunk.outDir, normalizedName)
 
       if (isIndex) {
         name = '.'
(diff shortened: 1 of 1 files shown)


src/features/pkg/exports.test.ts
```diff
@@ -462,6 +537,119 @@ describe.concurrent('generateExports', () => {
...
   })
+
+  test('generate css exports with css exports', async ({ expect }) => {
+    const results = generateExports(
+      FAKE_PACKAGE_JSON,
+      { es: [genChunk('index.js')] },
+      {
+        exports: {},
+        logger: globalLogger,
+        css: resolveCssOptions({ splitting: false }),
```
sxzz, line 531: Should not be exported, if `splitting` is disabled, but no css chunks are existed.

jinghaihan: I added a check to verify whether a CSS chunk matching the `cssOptions.fileName` actually exists in `chunks` when `splitting === false`, before adding it to exports.
Related unit tests have been added to cover this case.

src/features/pkg/exports.ts
```diff
@@ -281,6 +285,34 @@ function exportMeta(
...
 }
 
+function exportCss(
+  exports: Record<string, any>,
+  chunks: ChunksByFormat,
+  pkgRoot: string,
+  outDir: string,
+  { splitting, fileName }: Required<CssOptions>,
+) {
+  if (splitting === false) {
```
sxzz:
```suggestion
  if (!splitting) {
```

src/features/pkg/exports.ts
```diff
@@ -281,6 +285,34 @@ function exportMeta(
...
+    const cssFileName = fileName
```
sxzz: Moving renaming to line 293

src/features/pkg/exports.ts
```diff
@@ -281,6 +285,34 @@ function exportMeta(
...
+
+    for (const chunksByFormat of Object.values(chunks)) {
+      for (const chunk of chunksByFormat) {
+        if (
+          chunk.type === 'asset' &&
+          chunk.outDir === outDir &&
```
sxzz: Remove `outDir` param

sxzz (approved):

PR #685, opened 2026-01-02
feat!: disable types when legacy isn't passed as well

lishaduck:

### Description

Disables `types` when using `exports`.

### Additional context

Made this in the GitHub editor, test snapshots will likely need updating.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "./package.json": "./package.json",
     "./client": "./client.d.ts"
   },
+  "types": "./dist/index.d.mts",
   "typesVersions": {
     "*": {
       "*": [
--- a/packages/create-tsdown/package.json
+++ b/packages/create-tsdown/package.json
@@ -20,6 +20,7 @@
     "./run": "./dist/run.mjs",
     "./package.json": "./package.json"
   },
+  "types": "./dist/index.d.mts",
   "typesVersions": {
     "*": {
       "*": [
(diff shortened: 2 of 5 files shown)


src/features/pkg/exports.ts
```diff
@@ -261,7 +261,7 @@ export async function generateExports(
   return {
     main: legacy ? main || module || pkg.main : undefined,
     module: legacy ? module || pkg.module : undefined,
-    types: cjsTypes || esmTypes || pkg.types,
+    types: legacy ? cjsTypes || esmTypes || pkg.types : undefined,
```
lishaduck: I think the dance we did in the above packages was good and right, but if you disagree, we could also do:

```suggestion
    main: legacy ? main || module || pkg.main : pkg.main,
    module: legacy ? module || pkg.module : pkg.module,
    types: legacy ? cjsTypes || esmTypes || pkg.types : pkg.types,
```

I don't love that though, it feels a bit too magical and means we don't auto-migrate users anymore.

sxzz:
This may introduce potential issues:
- Automatic detection of the dts option will no longer function.
- While `@vitejs/plugin-vue` provides type declarations, they are not shown on npmjs.com because the site depends on the `types` field, which is missing from `@vitejs/plugin-vue/package.json`.

As a result, we will not be merging this PR at this time. For further discussion, please visit https://github.com/publint/publint/issues/24

src/features/pkg/exports.ts
```diff
@@ -258,7 +261,7 @@ export async function generateExports(
...
```
sxzz:
```suggestion
    types: legacy ? cjsTypes || esmTypes || pkg.types : pkg.types,
```

sxzz (approved):
I prefer to keep the `types` field for tsdown itself, but I agree that we should no longer auto-fill the `types` field if it was not present originally. This change will not affect the default value of the `dts` option.

PR #746, opened 2026-02-01
fix: throw error when skipNodeModulesBundle and noExternal are used together

### LINKED ISSUE #741: When `skipNodeModulesBundle` and `iife format`, transitive dependencies in noExternal list is not be included
## relate 
https://github.com/vuejs/router/issues/2608

# repo
https://stackblitz.com/github/Gehbt/tsdown-bug-repro-iife-skipNodeModulesBundle

# reproduct
- run `pnpm run build`
- lookup `dist/vue-router.global.js`

#### Package dependency:
```mermaid
flowchart TD
    A(["@vue/devtools-api"]) --> B(["@vue/devtools-kit"])
```

#### Options(noExternal):

Image

### What is expected?

"@vue/devtools-api" and its dependencies  should be include

skipNodeModulesBundle should not skip the transitive dependencies of noExternal dependencies

#### Output:

Image 

`_vue_devtools_kit` is re-export by `@vue/devtools-api`

### Any additional comments?

Or emit some warning for this case yet?

sxzz:
## Summary

- Throw a `TypeError` when `skipNodeModulesBundle` and `noExternal` are both configured, as they are mutually exclusive options

## Context

`skipNodeModulesBundle` means "do not bundle any dependencies from node_modules", while `noExternal` means "force certain dependencies to be bundled". Using them together leads to ambiguous behavior where transitive dependencies of `noExternal` packages are still externalized by `skipNodeModulesBundle`.

Closes #741

### DIFF
--- a/src/config/options.ts
+++ b/src/config/options.ts
@@ -242,6 +242,11 @@ export async function resolveUserConfig(
     const noExternalPatterns = toArray(noExternal)
     noExternal = (id) => matchPattern(id, noExternalPatterns)
   }
+  if (skipNodeModulesBundle && noExternal != null) {
+    throw new TypeError(
+      '`skipNodeModulesBundle` and `noExternal` are mutually exclusive options and cannot be used together.',
+    )
+  }
   if (inlineOnly != null && inlineOnly !== false) {
     inlineOnly = toArray(inlineOnly)
   }


dackmin:
Hey, I think this change is kind of breaking!
I was under the assumption that it meant "skip any dependency EXCEPT these ones" and it worked perfectly until this PR.
Do we have any replacement to achieve this? `inlineOnly` doesn't seem to do anything when `skipNodeModulesBundle` is enabled either, and putting all the dependencies in `external` does not include transitive dependencies and breaks the build altogether.
Thanks!

sxzz:
@dackmin Please open a new issue describing the problem you encountered in detail, and provide a minimal reproduction.

PR #769, opened 2026-02-11
feat(exports): support object option for `customExports`

### LINKED ISSUE #767: `exports.customExports` should accept an object

I believe `exports.customExports` should support an object as argument in addition to a function.

It feels a bit excessive to have to use a function to define one or two extra exports.

### Suggested solution

```ts
import { defineConfig } from "tsdown";

export default defineConfig({
    entry: "./src/index.ts",
    copy: ["./src/binding/worker.wasm"],
    exports: {
        customExports: {
            "./worker.wasm": "./dist/worker.wasm",
        },
    },
});
```

### Alternative

In my specific use-case it could also be possible to allow any files that were copied using `copy` to have exports generated.

### Additional context

_No response_

### Validations

Joery-M:

(I did AI generate the Chinese translation of the docs change, the rest is all written by me)

### Description

This PR adds support for providing an object directly to `exports.customExports` instead of having to make a function.

### Linked Issues

Closes #767

### Additional context

I modified the root tsdown.config.ts to use this change since it is exactly the use-case this PR is trying to improve.


src/features/pkg/exports.ts
```diff
@@ -15,7 +15,25 @@ import type {
   RolldownCodeChunk,
 } from '../../utils/chunks.ts'
 import type { Awaitable } from '../../utils/types.ts'
-import type { PackageJson } from 'pkg-types'
+import type { PackageJson, PackageJsonExports } from 'pkg-types'
+
+// Get the type `PackageJsonExportsObject` from pkg-types by extracting it from the `PackageJsonExports` union.
+// If in the future `PackageJsonExportsObject` is exported by pkg-types, this can be removed.
+type PackageJsonExportsObject = Extract<
```
sxzz: I prefer to keep using `Record<string, any>`

Joery-M: I liked the idea for intellisense, but I agree that the implementation is iffy. Fixed.

PR #773, opened 2026-02-19
fix(node-protocol): respect alias and tsconfig paths when stripping node: prefix

### LINKED ISSUE #772: `nodeProtocol: "strip"` ignores `paths` resolution for stripped `node:` imports

https://github.com/lazuee-dev/repro-tsdown-paths-nodeProtocol-strip

### What is expected?

```ts
Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });

//#region src/crypto-polyfill.ts
var _ref, _ref$crypto, _crypto, _crypto$randomUUID;
function hex(byte) {
	return byte.toString(16).padStart(2, "0");
}
(_ref$crypto = (_ref = globalThis).crypto) !== null && _ref$crypto !== void 0 || (_ref.crypto = {});
(_crypto$randomUUID = (_crypto = globalThis.crypto).randomUUID) !== null && _crypto$randomUUID !== void 0 || (_crypto.randomUUID = function randomUUID() {
	const bytes = new Uint8Array(16);
	for (let i = 0; i < 16; i++) bytes[i] = Math.trunc(Math.random() * 256);
	return hex(bytes[0]) + hex(bytes[1]) + hex(bytes[2]) + hex(bytes[3]) + "-" + hex(bytes[4]) + hex(bytes[5]) + "-" + hex(bytes[6]) + hex(bytes[7]) + "-" + hex(bytes[8]) + hex(bytes[9]) + "-" + hex(bytes[10]) + hex(bytes[11]) + hex(bytes[12]) + hex(bytes[13]) + hex(bytes[14]) + hex(bytes[15]);
});
const randomUUID = globalThis.crypto.randomUUID.bind(globalThis.crypto);
var crypto_polyfill_default = globalThis.crypto;

//#endregion
exports.randomUUID = randomUUID;
```

```js
Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
let crypto = require("crypto");

exports.randomUUID = crypto.randomUUID;
```

### Any additional comments?

```sh
bun run build
```

lazuee:
### Description

When `nodeProtocol: "strip"` is configured, the `NodeProtocolPlugin` runs with `order: 'pre'` and immediately marks stripped `node:` imports as `external: true`. This short-circuits the entire resolution chain, so `alias` and tsconfig `paths` configurations that map the stripped module name (e.g., `crypto`) to a polyfill are never consulted.

The fix changes the `strip` handler to call `this.resolve()` with the stripped ID before marking it as external. If another resolver (alias, tsconfig paths) resolves it to a non-external module, that resolution is used instead.

### Linked Issues

Fixes #772

### Additional context

The test covers both resolution methods in a single case:
- `alias`: `node:path` → stripped to `path` → resolved via `alias` to a polyfill
- `tsconfig paths`: `node:crypto` → stripped to `crypto` → resolved via `paths` to a polyfill

### DIFF
--- a/tests/issues.test.ts
+++ b/tests/issues.test.ts
@@ -174,6 +174,26 @@ describe('issues', () => {
     })
   })
 
+  test.fails('#668', async (context) => {
+    const { outputFiles, fileMap } = await testBuild({
+      context,
+      files: {
+        'shared.css': `.class-shared { color: red; }`,
+        'entry1.css': `@import './shared.css'; .class-entry1 { color: red; }`,
+        'entry2.css': `@import './shared.css'; .class-entry2 { color: red; }`,
+      },
+      options: {
+        entry: ['entry1.css', 'entry2.css'],
+      },
+    })
+    expect(outputFiles).toContain('entry1.css')
+    expect(outputFiles).toContain('entry2.css')
+    expect(fileMap['entry1.css']).toContain('class-entry1')
+    expect(fileMap['entry2.css']).toContain('class-entry2')
+    expect(fileMap['entry1.css']).toContain('class-shared')
+    expect(fileMap['entry2.css']).toContain('class-shared')
+  })
+
   test('#772', async (context) => {
     const { fileMap, outputFiles } = await testBuild({
       context,
@@ -198,25 +218,5 @@ describe('issues', () => {
     expect(fileMap['index.mjs']).not.toMatch(/from ['"]path['"]/)
     expect(fileMap['index.mjs']).toContain('polyfill-uuid')
     expect(fileMap['index.mjs']).not.toMatch(/from ['"]crypto['"]/)
-  })
-
-  test.fails('#668', async (context) => {
-    const { outputFiles, fileMap } = await testBuild({
-      context,
-      files: {
-        'shared.css': `.class-shared { color: red; }`,
-        'entry1.css': `@import './shared.css'; .class-entry1 { color: red; }`,
-        'entry2.css': `@import './shared.css'; .class-entry2 { color: red; }`,
-      },
-      options: {
-        entry: ['entry1.css', 'entry2.css'],
-      },
-    })
-    expect(outputFiles).toContain('entry1.css')
-    expect(outputFiles).toContain('entry2.css')
-    expect(fileMap['entry1.css']).toContain('class-entry1')
-    expect(fileMap['entry2.css']).toContain('class-entry2')
-    expect(fileMap['entry1.css']).toContain('class-shared')
-    expect(fileMap['entry2.css']).toContain('class-shared')
   })
 })


lazuee:
Using `https://pkg.pr.new/tsdown@773` with an [`alias`](https://github.com/lazuee-dev/repro-tsdown-paths-nodeProtocol-strip/commit/667a688), `this.resolve()` returns `null` unexpectedly,
tsconfig [`paths`](https://github.com/lazuee-dev/repro-tsdown-paths-nodeProtocol-strip/commit/c9719aa) works as expected tho,. And also doing `pnpm test` with this PR, `test#772` works as expected.

Idk, why `this.resolve()` returns `null` when using `alias` 🤔 
https://github.com/lazuee-dev/repro-tsdown-paths-nodeProtocol-strip

sxzz:
@lazuee Your `alias` option is incorrect:

```js
alias: {
	"crypto": "./src/crypto-polyfill.ts",
},
```

```js
// src/index.ts
import 'crypto'
// rewrite to
import './src/crypto-polyfill.ts' // which redirect to src/src/crypto-polyfill.ts
```

You can use absolute path or change it to `./crypto-polyfill.ts`.

lazuee:
@sxzz It works, but why does the relative path `./src/crypto-polyfill.ts` resolve to `src/src/crypto-polyfill.ts`? Is the `alias` root directory based on the current directory of the `entry` file, not on the root directory?

on the [rolldown#resolve.alias](https://rolldown.rs/reference/InputOptions.resolve#alias) it seems it based on the root directory.

sxzz:
It's by design in Rolldown.

resolve.alias will not call [resolveId](https://rolldown.rs/reference/Interface.Plugin#resolveid) hooks of other plugin. If you want to call resolveId hooks of other plugin, use viteAliasPlugin from rolldown/experimental instead. You could find more discussion in [this issue](https://github.com/rolldown/rolldown/issues/3615)

https://repl.rolldown.rs/#eNptUb1qwzAQfpVDi1twnN3QqaXQpVs2L4p1TgTyyUinNMH43aufJLhpwFjS3X0/+jSLQbSzcNYYZX+o6S0N+tCwT1US7bNOLfrY0ONkHcMMCgdN+J7bsMDg7AjVDVZ11BGe82gclMHk9Q54mTsC0DQFbqFQNiOybHKphu0W+IgwaIMePBrsGRXsL7m6D8yWIH4o+yOw3CeyyUgerBtbqMgqrOpUdOitOWELWQ9AGi39/QTQu8vENkKabdluJmsuUdbEGxeKhzHv+qejyfK3Zfi0gVTBLRmf/strR2KpRQJrUnheJf2n9pCxk6TsuNt9fdwTLtrrfOMLeQat4G01/5IUa4GiZRfwKv3f99rEk26xc9UZAvWsY+prlRY8O02HEqlDDo6gunFsQtAqel3S9aOJU2SLz4SexfIL+MHcaA==

lazuee:
I see, thanks for the info!

PR #778, opened 2026-02-24
fix(copy): avoid unnecessary warning when copy options are empty

### LINKED ISSUE #774: Empty `copy` options should not warn `No files matched for copying`

https://stackblitz.com/edit/github-ggwk8zxw?file=tsdown.config.ts

Run `pnpm build`.

### What is expected?

With `copy.from: []`, `copy: []`, or a function that returns an empty array, I'd expect the `No files matched for copying` warning not to show.

A `No files matched for copying` warning is logged, which subsequently fails on CI.

### Any additional comments?

While it doesn't make sense to configure an literal empty array as in the example, if a `copy` function conditionally returns an empty array, I'd expect the warning not to be logged in that case.

Basically, I want to conditionally copy files if they are present, otherwise don't copy any files. My use case is in a monorepo with a top-level tsdown config shared by many packages which may or may not contain assets I want to copy. Statically declaring the copy config for each package ahead of time _would_ work, but is quite tedious, so I'd prefer to glob for them ahead of time, and if any are found, tell tsdown to copy them to `dist`.

rzzf:

### Linked Issues

fix #774

### DIFF
--- a/src/features/copy.ts
+++ b/src/features/copy.ts
@@ -66,10 +66,7 @@ export async function copy(options: ResolvedConfig): Promise<void> {
     )
   ).flat()
 
-  if (!resolved.length) {
-    options.logger.warn(options.nameLabel, `No files matched for copying.`)
-    return
-  }
+  if (!resolved.length) return
 
   await Promise.all(
     resolved.map(({ from, to, verbose }) => {


sxzz:
This is not a correct fix.

PR #779, opened 2026-02-24
fix(css): not generating sourcemap

### LINKED ISSUE #753: Sourcemap not generated when css.splitting is false

https://stackblitz.com/edit/github-6nqaq67o?file=tsdown.config.ts

```
npm run build
```

### What is expected?

When `sourcemap: true` is configured, `.js.map` files should be generated in the output directory alongside the JavaScript files, regardless of the `css.splitting` option.

When `css.splitting` is set to `false`, sourcemap files are not generated.

## Analysis

I confirmed through logging that `cssAssets` incorrectly contains `.js.map` files.

Looking at the code, line 29 in `src/features/css/splitting.ts` uses `fileName` (from `config.css.fileName`) instead of `asset.fileName` in the condition:

https://github.com/rolldown/tsdown/blob/e3db8764f646f28e651a6183090b1d0d3f4a135f/src/features/css/splitting.ts#L29

## Suggested Fix (not confirmed)

This is my assumption based on the code analysis. The actual root cause may differ.

```diff
- RE_CSS.test(fileName)
+ RE_CSS.test(asset.fileName)

H4ad:
### Description

Basically applies the fix suggested at #753, I run locally and fixed the issue to emit the sourcemap for CSS files.

### Linked Issues

Fixes #753

### DIFF
--- a/src/features/css/splitting.ts
+++ b/src/features/css/splitting.ts
@@ -27,7 +27,7 @@ export function CssCodeSplitPlugin(
         if (
           asset.type === 'asset' &&
           typeof asset.source === 'string' &&
-          RE_CSS.test(fileName)
+          RE_CSS.test(asset.fileName)
         ) {
           cssAssets.set(asset.fileName, asset.source)
         }


H4ad:
Actually, this does not fix it completly, the source map was emitted but with the wrong name (expected style.css.map but was <module>.css.map)

sxzz:
The CSS-related implementation has been removed from Rolldown and will undergo a significant rewrite.

https://github.com/rolldown/rolldown/pull/8399

PR #780, opened 2026-02-25
fix(copy): downgrade "no files matched" from warn to info

adamlohner:
## Summary

Fixes https://github.com/rolldown/tsdown/issues/774

When the `copy` option is configured with a glob pattern that matches zero files (e.g., `copy: [{ from: 'src/**/*.png', to: 'dist' }]`), the build fails in CI because `failOnWarn` defaults to `'ci-only'`, which promotes the warning to an error via `process.exitCode = 1`.

An empty glob match for copy is a benign, expected condition — not a build problem — so this downgrades the log level from `warn` to `info` to prevent it from failing the build.

### DIFF
--- a/src/features/copy.ts
+++ b/src/features/copy.ts
@@ -69,7 +69,7 @@ export async function copy(options: ResolvedConfig): Promise<void> {
   ).flat()
 
   if (!resolved.length) {
-    options.logger.info(options.nameLabel, `No files matched for copying.`)
+    options.logger.warn(options.nameLabel, `No files matched for copying.`)
     return
   }
 


adamlohner:
hey @sxzz, saw your comment https://github.com/rolldown/tsdown/pull/778#issuecomment-3957295146 here, hoping this is an acceptable fix. if not, can you share more details on how you'd like to see this addressed? thanks!

sxzz (commented):
Essentially, we want to follow the description in the original issue: if no files are specified for copying, a warning should not be reported. 

As for the concern that warnings might cause failures in CI environments (`failOnWarn`), this option has already been set to default to false in all cases in the next release.

PR #819, opened 2026-03-10
feat: rename onlyAllowBundle to onlyBundle

peaklabs-dev:

### Description

- rename `deps.onlyAllowBundle` to `deps.onlyBundle` as it aligns better with the other options `deps.neverBundle`, `deps.alwaysBundle`, is shorter and also reads more naturally. `deps.onlyBundle=false` to suppress all warning still reads clean so it should work in all cases.

### DIFF
--- a/src/features/deps.ts
+++ b/src/features/deps.ts
@@ -53,7 +53,7 @@ export interface DepsConfig {
    */
   onlyBundle?: Arrayable<string | RegExp> | false
   /**
-   * @deprecated Use `deps.onlyBundle` instead.
+   * @deprecated Use {@link onlyBundle} instead.
    */
   onlyAllowBundle?: Arrayable<string | RegExp> | false
   /**
--- a/docs/options/dependencies.md
+++ b/docs/options/dependencies.md
@@ -142,7 +142,7 @@ The following top-level options are deprecated. Please migrate to the `deps` nam
 | `external`              | `deps.neverBundle`           |
 | `noExternal`            | `deps.alwaysBundle`          |
 | `inlineOnly`            | `deps.onlyBundle`            |
-| `onlyAllowBundle`       | `deps.onlyBundle`            |
+| `deps.onlyAllowBundle`  | `deps.onlyBundle`            |
 | `skipNodeModulesBundle` | `deps.skipNodeModulesBundle` |
 
 ## Summary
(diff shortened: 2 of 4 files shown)


peaklabs-dev:
@sxzz Yes I am aware and I fully get you point but I think it is not that big of a deal as the "old" options will continue to work just with a warning and most people most likely have not even upgraded to v0.21 yet so they will not notice it and if they have, it takes like 5 seconds to update.

src/features/deps.ts
```diff
@@ -98,16 +102,25 @@ export function resolveDepsConfig(
...
+      '`deps.onlyAllowBundle` is deprecated. Use `deps.onlyBundle` instead.',
     )
-    onlyAllowBundle = config.inlineOnly
+    onlyBundle = config.deps.onlyAllowBundle
+  }
+  if (config.inlineOnly != null) {
+    if (onlyBundle != null) {
+      throw new TypeError(
+        '`inlineOnly` is deprecated. Cannot be used with `deps.onlyBundle`.',
+      )
```
silverwind, line 120: Why is a deprecation a error? Logging it should have been enough.

sxzz: What is your use case? Please provide your configuration.

Are you using `inlineOnly` and `onlyBundle` together? This is conflicting behavior.

silverwind: My use case is just to suppress noisy warnings like this when my intention is to bundle dependencies:

```
ERROR  Consider adding inlineOnly option to avoid unintended bundling of dependencies, or set inlineOnly: false to disable this warning.
```

- Initially I used `inlineOnly: false` to suppress
- Since v0.21.0 I use `onlyAllowBundle: false` to suppress
- Since v0.21.2 I use `onlyBundle: false` to suppress

Maybe consider removing this warning spam.

sxzz: The option has been renamed, so you need to remove the old one.

silverwind: Already done but I don't get why are you:

1. breaking builds via `throw` for a deprecation warning?
2. not labeling the change as breaking in the changelog?
3. pushing breaking changes in semver-patch releases?

I suggest you review such slop more closely 😉.

sxzz:
v0.21.2 does not contain any breaking changes. We only deprecated the old option and recommend using the new one.

The `onlyBundle` option did not exist before v0.21.2, so it could not have been in use. The `throw` only triggers if you upgrade to v0.21.2 **and** add `onlyBundle` while keeping the old deprecated option — i.e., using both at the same time. If you upgrade without making any changes to your config, nothing will break. That does not qualify as a breaking change.

Encountering an error from incorrect usage after an upgrade is not a breaking change.

PR #836, opened 2026-03-17
ci(workflows): run the workflow only in the rolldown/tsdown repository

wChenonly:

### Description

workflows only run in rolldown/tsdownrepository; otherwise, I fork the repository and often receive errors

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -1,6 +1,6 @@
 {
   "name": "@sxzz/tsdown",
-  "version": "0.21.3",
+  "version": "0.21.4",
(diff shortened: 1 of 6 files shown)


.github/workflows/unit-test.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   unit-test:
+    if: github.repository == 'rolldown/tsdown'
```
sxzz: Unit Test workflow should always run.

Issue #839, opened 2026-03-17
CSS modules - Can't use a class named "default"


StyleShit:
### Reproduction link or steps

https://github.com/StyleShit/repros/tree/main/tsdown-css-modules-default

### What is expected?

When I have this CSS, it should be generated as css module:

```css
// index.module.css

.default {
  color: red;
}
```

### What is actually happening?

I'm getting this error:

```
ERROR  Error: Build failed with 1 error:

[PARSE_ERROR] Error: Identifier expected. 'default' is a reserved word that cannot be used here.
   ╭─[ src/index.module.css:1:14 ]
   │
 1 │ export const default = "design-system-default";
   │              ───┬───  
   │                 ╰───── 
───╯
```

StyleShit:
I guess this is fixed by https://github.com/rolldown/tsdown/pull/840

PR #840, opened 2026-03-18
fix(css): use aliased exports for CSS module keys

wChenonly:

### Description

fix the issue where reserved words cannot be exported correctly in CSS modules

add a check for JavaScript reserved words to avoid using reserved words as named exports when exporting CSS modules

reserved words referenced 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words

### Linked Issues

https://github.com/rolldown/tsdown/issues/839

### DIFF
--- a/tests/css.test.ts
+++ b/tests/css.test.ts
@@ -1660,7 +1660,6 @@ describe('css', () => {
             modules: { generateScopedName: 'design-system-[local]' },
           },
         },
-        snapshot: true,
       })
 
       const js = fileMap['index.mjs']


packages/css/src/modules.ts
```diff
@@ -1,11 +1,88 @@
...
+function canUseNamedExport(key: string): boolean {
+  return VALID_ID_RE.test(key) && !RESERVED_WORDS.has(key)
+}
 
 export function modulesToEsm(modules: Record<string, string>): string {
   const lines: string[] = []
   for (const [key, value] of Object.entries(modules)) {
-    if (VALID_ID_RE.test(key)) {
+    if (canUseNamedExport(key)) {
       lines.push(`export const ${key} = ${JSON.stringify(value)};`)
```
sxzz: Use this pattern:
```
const _key = ...
export { _key as "arbitraryKey" }
```

packages/css/src/modules.ts
```diff
@@ -1,13 +1,17 @@
...
-
 export function modulesToEsm(modules: Record<string, string>): string {
   const lines: string[] = []
+  let index = 0
   for (const [key, value] of Object.entries(modules)) {
-    if (VALID_ID_RE.test(key)) {
-      lines.push(`export const ${key} = ${JSON.stringify(value)};`)
+    if (key === 'default') {
+      continue
     }
```
sxzz, line 9: Why?

tests/css.test.ts
```diff
@@ -1647,5 +1648,29 @@ describe('css', () => {
...
+        files: {
+          'index.ts': `import styles from './app.module.css'\nconsole.log(styles.default)`,
+          'app.module.css': `.default { color: red }`,
+        },
+        options: {
+          css: {
+            modules: { generateScopedName: 'design-system-[local]' },
+          },
+        },
+        snapshot: false,
```
sxzz: Enable snapshot.

tests/css.test.ts
```diff
@@ -1647,5 +1648,29 @@ describe('css', () => {
...
+      })
+
+      const js = await readFile(`${outputDir}/index.mjs`, 'utf8')
+      const css = await readFile(`${outputDir}/style.css`, 'utf8')
```
sxzz: This is clearly AI vibe-coded. Please be honest and check the checkbox on the PR template accordingly.

wChenonly: I am very sure that I did not use it, the only PR I used has been marked

wChenonly: I didn't refer to the original CSS for testing. test.ts, I also have tests in my own project, and I've done similar unified test entry points, something like this form, so I just migrated it directly.

packages/css/src/modules.ts
wChenonly: because 

[PARSE_ERROR] Error: A module cannot have multiple default exports.
   ╭─[ app.module.css:2:19 ]
   │
 2 │ export { _key0 as "default" };
   │                   ────┬────  
   │                       ╰────── 
 3 │ export default {"default":"design-system-default"};
   │        ───┬───  
   │           ╰───── 
───╯

Issue #843, opened 2026-03-19
`clean` is not respected on `format.outDir`


dmurvihill:
### Reproduction link or steps

```ts
import { defineConfig } from "tsdown";

export default defineConfig({
  clean: true,
  format: {
    "esm": {
      outDir: 'esm/',
    },
    "cjs": {
      outDir: 'cjs/',
    },
  }
});
```

### What is expected?

Put some junk in the `esm/` and `cjs/` folders and run tsdown. It should be cleaned up.

### What is actually happening?

It doesn't get cleaned up.

PR #847, opened 2026-03-20
fix(clean): fix the cleanup issue for a specific outDir format

wChenonly:

### Linked Issues
fix: https://github.com/rolldown/tsdown/issues/843

### DIFF
--- a/src/config/options.ts
+++ b/src/config/options.ts
@@ -92,6 +92,7 @@ export async function resolveUserConfig(
     write = true,
     exe = false,
   } = userConfig
+  const rawClean = clean
 
   const pkg = await readPackageJson(cwd)
   if (workspace) {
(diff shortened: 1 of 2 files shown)


sxzz:
Thanks for the PR, but I'm going to close this for a few reasons:

1. `resolveClean` is already called once at L135 for the global `clean`. This PR adds a second call inside the format loop using `rawClean`, which works but is roundabout. A cleaner approach would be to defer `resolveClean` into the format loop entirely rather than resolving twice.

2. `overrides?.outDir` is resolved and then spread again via `...overrides`, which overwrites `outDir` before it gets re-overwritten by `outDir: resolvedOutDir`. The intent is unclear and fragile.

3. The AI checkbox is unchecked, but this looks AI-generated. Please review our contributing guidelines.

I'll fix this separately. The issue itself is valid though — thanks for reporting #843.

Issue #848, opened 2026-03-20
scss is not compiled correctly in vue sfc


LateDreamXD:
### Reproduction link or steps

```jsonc
// package.json
{
  // ...,
  "dependencies": {
    "vue": "^3.5.30"
  },
  "devDependencies": {
    // ...,
    "@tsdown/css": "^0.21.4",
    "sass-embedded": "^1.98.0",
    "tsdown": "^0.21.4",
    "typescript": "^5.9.3",
    "unplugin-vue": "^7.1.1",
    "vue-tsc": "^3.2.6"
  }
}
```

```ts
// tsdown.config.ts
// ...;

export default defineConfig({
	deps: {
		alwaysBundle: ['vue']
	},
	entry: { content: 'src/content/index.ts' },
	plugins: [vue()],
	css: {
		fileName: 'content.css',
		target: 'chrome140'
	},
	platform: 'browser',
	minify: true,
	format: 'iife',
	tsconfig: 'tsconfig.app.json'
});

// src/content/index.ts
import { createApp } from 'vue';
import App from './App.vue';

createApp(App).mount('#app');
```
```vue
<!-- src/content/App.vue -->
<template><!-- ... --></template>

<style lang="scss" scoped>
@mixin shared {
  // shared
}
.el1 {
  @include shared;
  // el1
}
.el2 {
  @include shared;
  // el2
}
</style>
```

### What is expected?

```css
/** dist/content.css */
.el1 {
  /** shared */
  /** el1 */
}

.el2 {
  /** shared */
  /** el2 */
}
```

### What is actually happening?

```css
/** dist/content.css */
@mixin shared {
  /** shared */
}

@include shared;
@include shared;
```

wChenonly:
can you provide a minimal reproducible example?

LateDreamXD:
sure, here is: [LateDreamXD/demo](/LateDreamXD/demo)

wChenonly:
> sure, here is: [LateDreamXD/demo](/LateDreamXD/demo)

get，I am trying to fix a bug.

PR #849, opened 2026-03-20
fix(css): preprocess vue sfc style virtual modules by lang query

wChenonly:

### Description

Fix Vue SFC style virtual module handling in @tsdown/css.

Previously, Vue style modules such as App.vue?vue&type=style&lang.scss were identified only from their cleaned file path (App.vue). That caused @tsdown/css to miss the preprocessor language from the query string, so Sass/Less/Stylus preprocessing was skipped and the original syntax could leak into emitted CSS.

This change makes preprocessor detection query-aware for virtual style modules and ensures the preprocessor pipeline uses a normalized filename for downstream processing. It also adds regression coverage for Vue SFC styles using scss, less, and stylus.

### Linked Issues
https://github.com/rolldown/tsdown/issues/848

### Additional context
Main review focus:
1. query-based preprocessor detection for Vue SFC virtual style ids
2. filename normalization passed through the preprocessor/PostCSS pipeline
3. regression coverage in CSS tests for Vue SFC lang="scss", lang="less", and lang="stylus"


sxzz:
The issue is real, but I'll fix it separately.

The pattern here (issue filed by someone else → immediate large PR with extensive tests and boilerplate from the same contributor as #847) suggests this is vibe-coded without deep understanding of the codebase.

PR #755, opened 2026-02-04
fix: type only packages should be considered when marking module external

kalvenschraut:
- [x] This PR contains AI-generated code, **but I have carefully reviewed it myself**. Otherwise, my PR may be closed.

### Description

As I tried to make the tests show if a type only package like @types/json-schema is prod/peer dependency it should be considered external, but only if the runtime dep wouldn't be inlined due to being a dev dependency.
### Linked Issues
#745

Issue #745, opened 2026-02-01
Bundles types from `@types/unist` etc.


lishaduck:
### Reproduction link or steps

https://stackblitz.com/edit/github-cuq1c6zc-syns2mmt?file=tsdown.config.ts

### What is expected?

tsdown should externalize `@types/` dependencies

### What is actually happening?

tsdown is not externalizing `@types/` dependencies when there isn't an actual package.

### Any additional comments?

Found in https://github.com/flint-fyi/flint/pull/2213 (where this is catching it correctly, but I couldn't fix it sufficiently)

kalvenschraut:
I noticed this for @types/json-schema too and I think the solution _should_ depend on where you have the @types/json-schema declared as a dependency or at least that will be most consistent with non type only packages.

For instance if
- @types/json-schema exists as a peer-dependency / normal dependency then externalize
- @types/json-schema is dev dependency internalize the type
    -  needs to be internalized so libs types that depend on it are accessible with strict dep package managers like pnpm
- if a package is installed that has both @types/foo and foo installed then it should follow whatever type of dependency foo is
    - currently tsdown doesn't distinguish between the two since they have the same specifier to the code, so need to internalize both since the actual package is being internalized

sxzz:
In your case, the issue arises because the warning message states: `or set inlineOnly: false to disable this warning.`. Since `failOnWarn` defaults to true in CI environments, this warning causes the build to fail.

Read more at https://x.com/antfu7/status/2015629048228249869

lishaduck:
> Read more at [x.com/antfu7/status/2015629048228249869](https://x.com/antfu7/status/2015629048228249869)

I don't use Twitter and can't see the thread but I think I agree? Seems like he's encouraging putting types in `dependencies`, which is exactly what I'm doing.

> In your case, the issue arises because the warning message states: or set inlineOnly: false to disable this warning.. Since failOnWarn defaults to true in CI environments, this warning causes the build to fail.

But I don't want to bundle the types. They are provided as runtime dependencies (as they should be and Anthony seems to be encouraging).
I set `failOnWarn` so it errors for the repro. As you said, that's not what's causing the issue. It's that tsdown is trying to bundle when it _shouldn't_ because they're not in `devDependencies`. It should be externalizing them.

lishaduck:
On-rereading I decided maybe this was the encouraged pattern, but it didn't work either... 🤔

```ts
  inlineOnly: [],
  external: ['@types/mdast'],
```

[...]

PR #852, opened 2026-03-21
fix(deps): correctly resolve root `@types` packages for dts deep imports

brc-dd:

### Description

Follow up to 0be7c6d / PR #755 for issue #745.

PR #755 fixed externalizing type-only packages declared as `@types/*`, but it still derived the DefinitelyTyped package name from the full import specifier. That caused deep imports like `markdown-it/lib/token.mjs` to be checked as `@types/markdown-it__lib__token.mjs` instead of `@types/markdown-it`.

### DIFF
--- a/src/features/deps.ts
+++ b/src/features/deps.ts
@@ -218,7 +218,7 @@ export function DepsPlugin(
           if (chunk.type === 'asset') continue
 
           for (const id of chunk.moduleIds) {
-            const parsed = await parseBundledDep(id)
+            const parsed = await readBundledDepInfo(id)
             if (!parsed) continue
 
             deps.add(parsed.name)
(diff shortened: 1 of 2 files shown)


src/features/deps.ts
```diff
@@ -192,6 +186,24 @@ async function parseBundledDep(
   } catch {}
 }
 
+export function getPackageName(id: string): string | undefined {
+  const [first, second] = id.split('/', 3)
```
brillaintlcd: Setting the limit at 3 is also good for performance and adds up in large builds 👍

src/features/deps.ts
```diff
@@ -192,6 +186,24 @@ async function parseBundledDep(
...
+  return first
+}
+
+export function getTypesPackageName(id: string): string | undefined {
+  const packageName = getPackageName(id)
+  if (!packageName) return
+
+  return `@types/${packageName.replace(/^@/, '').replace('/', '__')}`
+}
+
```
brillaintlcd: This is also more robust and safe against malformed package names as well

verbose117 (approved):

PR #856, opened 2026-03-23
feat(dts): add `cjsReexport` option to eliminate dual module type hazard

mandarini:
# Description

When building dual ESM+CJS formats, tsdown runs two separate TypeScript compilation passes, one producing `.d.mts`, one producing an independent `.d.cts`. Because TypeScript's class identity is nominal, these two declarations are treated as unrelated types by consumers that mix CJS and ESM resolution paths (e.g. NestJS apps importing a package that also has ESM-only internal dependencies). This produces `TS2352` "neither type sufficiently overlaps" when casting between types derived from the same class.

This PR adds a `dts.cjsReexport: boolean` option (opt-in, default `false`). When enabled in a dual-format build, tsdown skips the second compilation pass and instead emits a `.d.cts` stub via a Rolldown plugin (`generateBundle` hook) that re-exports from the corresponding `.d.mts`:

```js
// dist/index.d.cts (generated)
export * from './index.d.mts'
```

Both CJS and ESM consumers now resolve to the same declaration, eliminating the dual module type hazard entirely.

> Note: `export *` does not re-export default. Packages with a default export and named exports should be aware of this — for named-export-only packages this is a non-issue.

## Implementation

The stub is emitted by `CjsDtsReexportPlugin` (in `src/features/rolldown.ts`), which uses Rolldown's `generateBundle` hook to call `this.emitFile`. This integrates naturally with the build pipeline and respects the `write` option without any direct filesystem access.

## Additional info

This PR contains AI-assisted code, but I have carefully reviewed it myself. We use `tsdown` on https://github.com/supabase/supabase-js and it would be nice if we could use this.

### DIFF
--- a/dts.snapshot.json
+++ b/dts.snapshot.json
@@ -127,6 +127,7 @@
     "DepsConfig": "interface DepsConfig {\n neverBundle?: ExternalOption\n alwaysBundle?: Arrayable<string | RegExp> | NoExternalFn\n onlyBundle?: Arrayable<string | RegExp> | false\n onlyAllowBundle?: Arrayable<string | RegExp> | false\n skipNodeModulesBundle?: boolean\n}",
     "DepsPlugin": "declare function DepsPlugin(_: ResolvedConfig, _: TsdownBundle): Plugin",
     "DevtoolsOptions": "interface DevtoolsOptions extends NonNullable<InputOptions['devtools']> {\n ui?: boolean | Partial<StartOptions>\n clean?: boolean\n}",
+    "DtsOptions": "interface DtsOptions extends Options$1 {\n cjsReexport?: boolean\n}",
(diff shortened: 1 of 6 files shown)


sxzz:
Thanks for the PR! The use case is valid — the dual module type hazard is a real pain point.

However, the current implementation writes the `.d.cts` stub directly to the filesystem via `fs.writeFile`. This should be implemented as a Rolldown plugin instead (e.g. using the `generateBundle` hook to emit the reexport stub), so it integrates properly with the build pipeline and respects the `write` option naturally.

Could you refactor the approach to use a Rolldown plugin?

mandarini:
Thanks for the review @sxzz ! Sure, I'll do the refactor!

mandarini:
@sxzz updated!

sxzz (approved):

mortyccp:
There is error in `dist/index.d.cts`
> A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file './index.mts' instead?ts(2846)

I think should be related to this https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax
And updating to `export type * from './index.d.mts'` resolve the issue.

sxzz:
Fixed via 8ecd1b5a701e8a6f4bc68b095c41a6373e6e2f52

PR #876, opened 2026-03-29
docs: use `@my-org/source` pattern as custom condition

ocavue:

### Description

As a library author, using common words like `dev` or `development` as a custom condition name is generally a bad practice. Your library's `package.json` will contain this custom condition in its `exports` field. If your users also happen to use `dev` as a custom condition in their own project, they may inadvertently resolve to your raw TypeScript source code instead of the compiled JavaScript output. Resolving to raw `.ts` source files can lead to unpredictable behavior due to differences in the consumer's TypeScript configuration, so distributing raw TypeScript source via commonly-named conditions should be avoided.

There is no strict convention for naming custom conditions. A common pattern I've observed is to use a scoped name like `@my-org/source`. This convention is used by both TypeScript itself and Nx. For example:

- [Nx docs: switch to workspaces project references](https://github.com/nrwl/nx/blob/c9253fba6707aab4362cd561c28f087dc0457022/astro-docs/src/content/docs/technologies/typescript/Guides/switch-to-workspaces-project-references.mdoc#L358)
- [`@typescript/source`](https://github.com/microsoft/typescript-go/blob/98545e9a34274cb61b4b521b8f49336e1ddff08a/_packages/api/package.json#L11)
- [`@monosize/source`](https://github.com/microsoft/monosize/blob/fe208796261480d973864e528e4f0452f8b11389/tsconfig.base.json#L23)

This PR renames the `dev` custom condition to `@tsdown/source` to follow this convention. I also update the documentation to encourage users to use `@my-org/source` pattern.


sxzz:
`devExports` will not exposed to user-land. `publishConfig#exports` will overwrite it.

ocavue:
`publishConfig#exports` doesn't work if you're _not_ using PNPM.

sxzz:
I think updating the documentation is fine, but we don't need to change our codebase, we're using pnpm.

ocavue:
@sxzz Done. Only docs and tests are updated. tsdown source code is still using `dev`.

PR #923, opened 2026-04-19
refactor(ansis): replace ansis dep with node api

### LINKED ISSUE #530: Replace `ansis` dependency with `util.styleText`
https://nodejs.org/api/util.html#utilstyletextformat-text-options

Available since Node v20.12.0 which matches the current minimum version of v20.19.0.

It seems the only complication is the `generateColor` function which currently returns an Ansis color specifier, but this type can be replaced with the type of the first parameter of `styleText`.

SoftAndoWeto:
### Description

This PR adds a util `styleText`. It also updates generated workspace label colors to use `styleText` formats instead of `ansis` color functions.
The `tsdown-migrate` package now uses `styleText` directly, and `ansis` has been removed from the package manifests, workspace catalog, and lockfile importers.

### Linked Issues

Closes #530


sxzz:
https://github.com/rolldown/tsdown/issues/530#issuecomment-3399154560

PR #960, opened 2026-05-25
fix: respect failOnWarn regardless of logger level

### LINKED ISSUE #952: `failOnWarn` is coupled to logging output; silent logging will succeed

This direct usage of `build` shows how when failOnWarn is true and logLevel is silent, build does not throw.
but when logLevel is warn, it throws as expected.

```ts
import {mkdtempSync, writeFileSync} from 'node:fs'
import {tmpdir} from 'node:os'
import {join} from 'node:path'
import {build} from 'tsdown'

// Create a temp project with a source file that imports a package not installed anywhere.
const dir = mkdtempSync(join(tmpdir(), 'tsdown-failonwarn-repro-'))
const entry = join(dir, 'index.ts')
writeFileSync(
  join(dir, 'package.json'),
  JSON.stringify({type: 'module', private: true}, null, 2),
)
writeFileSync(
  entry,
  `import express from 'express'\nexport const handler = () => express()\n`,
)

console.log('Test 1: failOnWarn=true, logLevel=silent')
try {
  await build({
    entry: {index: entry},
    outDir: join(dir, 'dist-silent'),
    format: 'esm',
    platform: 'node',
    dts: false,
    config: false,
    logLevel: 'silent',
    failOnWarn: true,
    deps: {alwaysBundle: () => true, onlyBundle: false},
  })
  console.log('  -> BUILD SUCCEEDED (bug: unresolved import was silently externalized)')
} catch (err) {
  console.log('  -> BUILD THREW:', err.message.split('\n')[0])
}

console.log('')

console.log('Test 2: failOnWarn=true, logLevel=warn')
try {
  await build({
    entry: {index: entry},
    outDir: join(dir, 'dist-warn'),
    format: 'esm',
    platform: 'node',
    dts: false,
    config: false,
    logLevel: 'warn',
    failOnWarn: true,
    deps: {alwaysBundle: () => true, onlyBundle: false},
  })
  console.log('  -> BUILD SUCCEEDED (unexpected)')
} catch (err) {
  console.log('  -> BUILD THREW:', err.message.split('\n')[0])
}
```

```
Test 1: failOnWarn=true, logLevel=silent
  -> BUILD SUCCEEDED (bug: unresolved import was silently externalized)

Test 2: failOnWarn=true, logLevel=warn
  -> BUILD THREW: Build failed with 1 error:
```

### What is expected?

I would expect both cases above to throw. Opting into `failOnWarn` feels like a warning, whether it is logged or not, should fail the build.

`build` succeeds when when logging is silent and failOnWarn is enabled.

### Any additional comments?

I acknowledge the option is not `failOnWarning`, but it's close enough that it leaves me wanting a simple way to fail all warnings even when logging is off.

spokodev:
## Problem

`failOnWarn: true` is documented as a way to fail a build whenever a warning is emitted. In practice it relied on rolldown's `defaultHandler` to escalate `'warn'` into a build error, but when the caller passes `logLevel: 'silent'` rolldown's `defaultHandler` becomes a no-op. So warnings such as unresolved imports get silently externalised and the build succeeds when the user explicitly opted into failing.

A common shape where this bites is CI scripts that pass `logLevel: 'silent'` to keep their own logs clean and `failOnWarn: true` to fail builds with unresolved imports. The current behaviour means those builds silently pass.

## Repro

```ts
import { mkdtempSync, writeFileSync } from 'node:fs'
import { tmpdir } from 'node:os'
import { join } from 'node:path'
import { build } from 'tsdown'

const dir = mkdtempSync(join(tmpdir(), 'tsdown-failonwarn-'))
writeFileSync(
  join(dir, 'package.json'),
  JSON.stringify({ type: 'module', private: true }),
)
writeFileSync(
  join(dir, 'index.ts'),
  `import express from 'express'\nexport const handler = () => express()\n`,
)

await build({
  entry: { index: join(dir, 'index.ts') },
  outDir: join(dir, 'dist'),
  format: 'esm',
  platform: 'node',
  dts: false,
  config: false,
  logLevel: 'silent',
  failOnWarn: true,
  deps: { alwaysBundle: () => true, onlyBundle: false },
})
// before: resolves, unresolved import was silently externalised
// after:  throws "Build failed with 1 error" as expected
```

The same call with `logLevel: 'warn'` already throws, so the bug only surfaces under silent.

## Fix

Decouple rolldown's `logLevel` from tsdown's `logger.level` whenever `failOnWarn` is enabled and the user-facing log level would otherwise suppress warnings:

```ts
logLevel: logger.options?.failOnWarn
  ? logger.level === 'silent' || logger.level === 'error'
    ? 'warn'
    : logger.level
  : logger.level === 'error'
    ? 'silent'
    : logger.level,
```

This keeps existing behaviour for callers who do not set `failOnWarn` and for callers whose `logLevel` already allows warnings. The escalation path inside the `onLog` handler is unchanged.

## Tests

Four direct-build tests added to `tests/e2e.test.ts` that bypass the `testBuild` helper, which is necessary because `testBuild`'s `inputOptions` override forces rolldown's `logLevel` to `'info'` and therefore could never reproduce the silent-mode path. The scenarios cover all four axes:

- **Positive**: silent + `failOnWarn` + warning => `build()` throws.
- **Negative**: silent + `failOnWarn` + clean build (no warnings) => `build()` succeeds.
- **Negative**: silent + `failOnWarn: false` + warning => `build()` succeeds (opt-out respected).
- **Regression preservation**: `warn` + `failOnWarn` + warning => `build()` still throws.

Without the fix the positive scenario fails on `main` (the build resolves silently); the other three pass on baseline already and continue to pass with the patch, confirming the fix narrowly targets the broken path.

`pnpm lint` clean, `pnpm typecheck` clean, full e2e suite green for all 5 `failOnWarn` cases (1 pre-existing + 4 new). The 3 pre-existing `exe.test.ts` failures on `main` are unrelated to this change.

Closes #952

## Note on disclosure

Bug surfacing and the repro are from `@tbeseda` in the issue body. I traced the coupling in `src/features/rolldown.ts`, reproduced it via the script above, applied the minimal log-level decoupling, and added the tests. AI assistance was used while drafting parts of the test boilerplate and this PR body; the fix and tests reflect my own engineering decisions.

### DIFF
--- a/src/features/rolldown.ts
+++ b/src/features/rolldown.ts
@@ -9,6 +9,7 @@ import {
   type ExternalOption,
   type ExternalOptionFunction,
   type InputOptions,
+  type LogLevelOption,
   type OutputOptions,
   type Plugin,
   type RolldownPluginOption,
@@ -110,6 +111,7 @@ async function resolveInputOptions(
     unused,
     watch,
   } = config
+  const loggerLevelIndex = LogLevels[logger.level]
 
   const plugins: RolldownPluginOption = []
 
@@ -173,7 +175,7 @@ async function resolveInputOptions(
     }
   }
 
-  if (report && LogLevels[logger.level] >= 3 /* info */) {
+  if (report && loggerLevelIndex >= 3 /* info */) {
     plugins.push(ReportPlugin(config, cjsDts, isDualFormat))
   }
 
@@ -230,6 +232,15 @@ async function resolveInputOptions(
     external = deps.neverBundle
   }
 
+  const logLevel: LogLevelOption =
+    // When failOnWarn is enabled, ensure rolldown's log level is at least
+    // 'warn' so its defaultHandler can escalate warnings to build errors.
+    logger.options?.failOnWarn && loggerLevelIndex < 2 /* warn */
+      ? 'warn'
+      : logger.level === 'error'
+        ? 'silent'
+        : logger.level
+
   const inputOptions = await mergeUserOptions(
     {
       input: entry,
@@ -251,17 +262,7 @@ async function resolveInputOptions(
         '.node': 'copy',
         ...loader,
       },
-      // When failOnWarn is enabled, ensure rolldown's log level is at least
-      // 'warn' so its defaultHandler can escalate warnings to build errors.
-      // Otherwise a caller passing logLevel: 'silent' would never fail because
-      // rolldown's defaultHandler becomes a no-op.
-      logLevel: logger.options?.failOnWarn
-        ? logger.level === 'silent' || logger.level === 'error'
-          ? 'warn'
-          : logger.level
-        : logger.level === 'error'
-          ? 'silent'
-          : logger.level,
+      logLevel,
       onLog(level, log, defaultHandler) {
         // suppress mixed export warnings if cjsDefault is enabled
         if (cjsDefault && log.code === 'MIXED_EXPORT') return
--- a/src/features/rolldown.ts
+++ b/src/features/rolldown.ts
@@ -251,7 +251,17 @@ async function resolveInputOptions(
         '.node': 'copy',
         ...loader,
       },
-      logLevel: logger.level === 'error' ? 'silent' : logger.level,
+      // When failOnWarn is enabled, ensure rolldown's log level is at least
+      // 'warn' so its defaultHandler can escalate warnings to build errors.
+      // Otherwise a caller passing logLevel: 'silent' would never fail because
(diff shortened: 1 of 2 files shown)


spokodev:
Gentle nudge on this one. It closes #952 and has been green for a few weeks, with an e2e test covering the failOnWarn behaviour. Could a maintainer take a look when there's a spare moment? I can rebase if it has drifted from main.

sxzz (approved):

PR #962, opened 2026-05-28
fix: hint dev-only deps in dts diagnostics

### LINKED ISSUE #743: When outputting undefined import messages, indicate that this could also happen due to a devDependencey that should by a normal dependency

I had to really debug to minimal reproduction to release that the issue why the latest tsdown started outputting a bunch of warnings on my was due to the type it stopped pulling in being a dev dependency and not a normal dependency.

### Suggested solution

Output a message that says this could happen because a dependency is marked a dev dependency instead of a prod dependency.

### Alternative

 Even better if tsdown can check that the required package that is missing  is a dev dependency and then output that info directly in this situation.

### Additional context

 This package is located inside a monorepo so it never actually caused problems that we noticed, but if I were to publish this package consumers of it would so I see the value of this warning. The main problem was that it was very time consuming to find this out

Image

reproduction repo/branch

https://github.com/RTVision/invalid-dts-generation/tree/invalid-type-gen

### Validations

M-Hassan-Raza:
Fixes #743.

This adds a declaration diagnostic hint when Rolldown reports a missing export from a package that only appears in devDependencies. The hint points users to move that package to dependencies or peerDependencies when published declarations need it.

The implementation keeps dependency ownership in deps.ts and declaration error handling in declaration-diagnostics.ts. It only handles structured MISSING_EXPORT diagnostics for declaration files, so it does not guess from unstructured messages or runtime errors.

Validated with pnpm typecheck, pnpm lint, pnpm test src/features/declaration-diagnostics.test.ts tests/e2e.test.ts --run, and pnpm test src/features/deps.test.ts --run.

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -14,6 +14,7 @@ import {
 import { warnLegacyCJS } from './features/cjs.ts'
 import { cleanChunks, cleanOutDir } from './features/clean.ts'
 import { copy } from './features/copy.ts'
+import { addDeclarationDiagnosticHints } from './features/declaration-diagnostics.ts'
 import { startDevtoolsUI } from './features/devtools.ts'
 import { isGlobEntry, toObjectEntry } from './features/entry.ts'
 import { buildExe } from './features/exe.ts'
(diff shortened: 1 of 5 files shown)


sxzz:
Thanks for the contribution. As noted in [this comment](https://github.com/rolldown/tsdown/issues/743#issuecomment-5036489112), the latest version of `rolldown-plugin-dts` already emits a sufficiently clear warning for this case, so I don't think this additional handling is necessary.

PR #969, opened 2026-06-05
ci: deploy docs to Void

Boshen:
## Summary
- add a Void deploy job for docs after unit tests on main
- install the Void CLI in the docs workspace
- update pnpm/actionspack generated files and build approvals for Void dependencies

## Verification
- pnpm exec actionspack verify
- pnpm -C docs run void --help
- git diff --check


sxzz:
Thanks for working on this.

I’d prefer that we don’t merge deploy/CI changes like this directly. Even if the change looks straightforward, it affects the production docs deployment path, so it should get at least one review from a tsdown maintainer before merging.

One issue here is that tsdown.dev is deployed from the `release` branch, not `main`, so the workflow didn’t match the actual docs deployment setup. I’ve fixed that in 742f3141333b237e33b2be15b34bf6b54439c876.

For future deploy-related changes, please leave the PR open until it has been reviewed by a maintainer.

PR #981, opened 2026-06-20
fix: bundle CSS from external dependencies instead of externalizing

### LINKED ISSUE #959: External CSS assets not being inlined when imported via `?inline`

minimal reproduction:  https://stackblitz.com/edit/github-2dddpu9g?file=src%2Findex.ts
- run `npm run build`
- examine output asset in`dist/index.js`
- notice:
  - the contents of `style.css` are inlined (as expected, this is correct)
  - the contents of `jodit/es2021/jodit.css` are not inlined (not expected, this is the bug)
  - Image

### What is expected?

I would expect when `?inline` is added to the import statement the contents of the file are inlined as a string and assigned to the import variable.

In `tsdown`, the import is being processed as a regular js import statement instead of an instruction to inline the css file as a string.

### Any additional comments?

I have prepared the exact same case in Vite 8 library mode (same source files, same tsconfig) to show how it functions in comparison.

https://stackblitz.com/edit/vitejs-vite-2v5kn8ki?file=src%2Findex.ts
- run `npm run build`
- examine output asset in`dist/index.js`
- notice:
  - the contents of `style.css` are inlined (as expected)
  - the contents of `jodit/es2021/jodit.css` are also inlined (as expected).
  - Image

compass-vin:
## Summary

Fixes #959

CSS files imported via JS/TS from production dependencies (listed in `dependencies`, `peerDependencies`, or `optionalDependencies`) were being marked as external by the deps plugin, which prevented the CSS plugin from processing them into the output bundle.

This affected both regular CSS imports and `?inline` CSS imports:
- `import 'jodit/es2021/jodit.css'` — CSS not included in output `style.css`
- `import style from 'jodit/es2021/jodit.css?inline'` — CSS not inlined as string

**Root cause:** `externalStrategy()` in `src/features/deps.ts` checked if an import's package name matched a production dependency and externalized it unconditionally. CSS files need to be bundled so the CSS plugin can process them.

**Fix:** Added a guard in `externalStrategy()` that returns `'no-external'` for CSS file imports (`.css`, `.scss`, `.sass`, `.less`, `.styl`, `.stylus`), stripping any query string (e.g. `?inline`) before testing. This runs after the `alwaysBundle` check but before the dependency externalization checks.

## Test plan

- [x] Added test: external dependency CSS is bundled via JS `import`
- [x] Added test: external dependency CSS is bundled via JS `?inline` import
- [x] All existing CSS tests pass (100 passed + 3 expected fail)
- [x] All existing deps tests pass (15 passed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/features/deps.ts
+++ b/src/features/deps.ts
@@ -3,7 +3,7 @@ import { isBuiltin } from 'node:module'
 import path from 'node:path'
 import { blue, underline, yellow } from 'ansis'
 import { createDebug } from 'obug'
-import { RE_DTS, RE_NODE_MODULES } from 'rolldown-plugin-dts/internal'
+import { RE_CSS, RE_DTS, RE_NODE_MODULES } from 'rolldown-plugin-dts/internal'
 import { and, id, importerId, include } from 'rolldown/filter'
 import {
   matchPattern,
@@ -328,6 +328,12 @@ export function DepsPlugin(
     const isDts = importer ? RE_DTS.test(importer) : false
     const alwaysBundle = (isDts && dts?.alwaysBundle) || jsAlwaysBundle
     if (alwaysBundle?.(id, importer)) {
+      return 'no-external'
+    }
+
+    // CSS files must always be bundled so the CSS plugin can process them
+    const cssCheckId = resolved?.id || id
+    if (RE_CSS.test(cssCheckId.split('?', 1)[0])) {
       return 'no-external'
     }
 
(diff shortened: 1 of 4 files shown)


compass-vin:
Tested fix is here: 

https://stackblitz.com/edit/github-2dddpu9g-22k8rjck?file=src%2Findex.ts

compass-vin:
@sxzz Hey, it's my first time contributing to this project so apologies if I'm missing a step. Is there anything I need to do to get this reviewed?

Thanks for all the work on this btw, I love this library.

PR #1016, opened 2026-07-13
docs: note about typescript@7.0.x to options/dts.md

bulebrainbrand:
document that `typescript@7.0.x` requires `@typescript/native-preview` when use dts generate to options/dts.md .

### Description

at moment, if you install only `typescript@7.0.x` package, you will fail d.ts file generating. so you must install also `@typescript/native-preview`. 
this Pull Request document that problem to options/dts.md.

Also, According to microsoft, `typescript^7.1.0` will have api. so this Pull Request touch only `typescript@7.0.x` .
if you think you should change expression, it's welcome. i'm not good of English.

### DIFF
--- a/docs/options/dts.md
+++ b/docs/options/dts.md
@@ -5,7 +5,7 @@ Declaration files (`.d.ts`) are an essential part of TypeScript libraries, provi
 `tsdown` makes it easy to generate and bundle declaration files for your library, ensuring a seamless developer experience for your users.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


sxzz:
That's not correct.

sxzz:
We supported both `typescript@7` and native-preview package.

bulebrainbrand:
umm, i will check again. if i'm misunderstanding, i'm so sorry.

bulebrainbrand:
actually it works. im so sorry. this issue is not correct.

PR #1025, opened 2026-07-19
ci: migrate Void deploy to GitHub OIDC

Boshen:
Authenticate the Void deploy via **GitHub OIDC** instead of a stored `VOID_TOKEN` secret. `void deploy` mints an OIDC token and exchanges it for a short-lived, project-scoped deploy token — no long-lived secret is stored.

## Changes
- Grant `id-token: write` on the `deploy-void` job.
- Drop `VOID_TOKEN` from the deploy step env (`VOID_PROJECT` unchanged).

Applied to both the actionspack source (`.github/workflows/src/deploy.yml`) and the generated `.github/workflows/deploy.yml`. I edited the generated file directly rather than running a full `actionspack` regen, so this doesn't pull unrelated upstream (`sxzz/workflows`) template changes; a future regen reproduces the same change from the source.

## Required follow-up (server-side, one-time)
OIDC only succeeds once the repo is connected to the `tsdown` project with the `github_actions` executor on Void:

```bash
void github connect tsdown \
  --repo rolldown/tsdown \
  --branch main \
  --executor github_actions \
  --workflow .github/workflows/deploy.yml
```

After the first successful OIDC deploy, the `VOID_TOKEN` secret can be removed.

### DIFF
--- a/.github/workflows/src/deploy.yml
+++ b/.github/workflows/src/deploy.yml
@@ -21,10 +21,11 @@ jobs:
     if: github.repository == 'rolldown/tsdown'
     runs-on: ubuntu-latest
     environment: void
+    permissions:
+      id-token: write
     steps:
       - uses: sxzz/workflows/setup-js@main
       - run: pnpm run docs:generate
       - run: pnpm -C docs run void deploy
         env:
-          VOID_TOKEN: ${{ secrets.VOID_TOKEN }}
           VOID_PROJECT: tsdown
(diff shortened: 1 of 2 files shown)


sxzz (approved):

.github/workflows/src/deploy.yml
```diff
@@ -21,10 +21,11 @@ jobs:
     if: github.repository == 'rolldown/tsdown'
     runs-on: ubuntu-latest
     environment: void
```
sxzz, line 23: `void` environment can be removed I guess?

PR #1031, opened 2026-07-25
docs: explain in-memory programmatic builds

ColumbusLabs:

### Description

Documents the Programmatic Usage write:false item from #247 across the English guide, Chinese guide, and synchronized tsdown skill reference.

The new section shows an in-memory build, explains chunk and asset content access, documents the current TsdownBundle[] return shape, notes remaining filesystem side effects, and calls out the watch-mode incompatibility. It also corrects the skill reference’s stale Promise<void> description.

This is AI-assisted documentation, not core code. The final diff was validated against the implementation with a live runtime probe and independently reviewed.

Validation:
- live build({ write:false, clean:false }) probe returned an in-memory bundle
- pnpm lint
- pnpm typecheck
- pnpm docs:generate
- pnpm docs:build
- targeted Prettier and git diff --check

### Linked Issues

Addresses the unchecked Programmatic Usage write:false documentation item in #247.

### Additional context

The docs explicitly avoid combining write:false with watch mode because current watch rebuilds do not expose returned bundles.

### DIFF
--- a/docs/advanced/programmatic-usage.md
+++ b/docs/advanced/programmatic-usage.md
@@ -16,4 +16,37 @@ await build({
 })
 ```
 
+## In-Memory Output
+
+Set `write: false` to access generated chunks and assets without writing the
+bundle output to disk:
+
+```ts twoslash
+import { build } from 'tsdown'
+
+const bundles = await build({
+  entry: ['src/index.ts'],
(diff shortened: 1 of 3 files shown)


sxzz (approved):

sxzz:
Thanks!

PR #1035, opened 2026-07-29
feat(css): support function form for `css.modules.localsConvention`

### LINKED ISSUE #1034: Support function form for `css.modules.localsConvention`

`tsdown` currently supports only string values for `css.modules.localsConvention`:

```ts
localsConvention?: 'camelCase' | 'camelCaseOnly' | 'dashes' | 'dashesOnly';
```

`Vite` supports a function form for `css.modules.localsConvention`, which allows projects to define custom CSS Modules export names:

```ts
css: {
  modules: {
    localsConvention: (className) =>
      className.replaceAll(/-([a-z0-9])/g, (_, char) => char.toUpperCase())
  }
}
```

This becomes an issue when migrating a project from Vite or another build setup that relies on custom CSS Modules key generation.

### Suggested solution

Allow css.modules.localsConvention to accept a function, similar to Vite:

```ts
localsConvention?:
  | 'camelCase'
  | 'camelCaseOnly'
  | 'dashes'
  | 'dashesOnly'
  | ((className: string) => string);
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

lazerg:

### Description

`css.modules.localsConvention` only accepted the four string presets, so projects migrating from Vite that pass a function to derive export names had no way to keep that behavior. This lets it also accept a function, matching Vite's signature `(originalClassName, generatedClassName, inputFile) => string`. The existing string conventions are unchanged.

### Linked Issues

Closes #1034

### DIFF
--- a/docs/options/css.md
+++ b/docs/options/css.md
@@ -386,6 +386,21 @@ Controls how class names are exported in JavaScript:
 | `'dashes'`        | `foo-bar` | `foo-bar`, `fooBar` |
 | `'dashesOnly'`    | `foo-bar` | `fooBar`            |
 
+A function can also be passed to derive the exported name from each class name:
+
+```ts
+export default defineConfig({
+  css: {
+    modules: {
+      localsConvention: (originalClassName, generatedClassName, inputFile) =>
+        originalClassName.replaceAll(/-([a-z0-9])/g, (_, char) =>
+          char.toUpperCase(),
+        ),
+    },
+  },
+})
+```
+
 ### `generateScopedName`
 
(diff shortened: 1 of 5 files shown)


sxzz:
Thanks for the PR! This feature has landed in main via b566f6d6f2d8fb6ab8b94cf1797da0525d56e566 with you credited as co-author.

I ended up implementing it with a slightly different approach on the PostCSS side: instead of applying the convention in tsdown's own `applyLocalsConvention`, the `localsConvention` option (including the function form) is now passed through to `postcss-modules` directly, which supports it natively with the same signature. This also fixes an issue where the user's `getJSON` callback was invoked twice on the PostCSS path (once with unconverted class names). The Lightning CSS path uses tsdown's own implementation as before, extended to support the function form as in your PR.

Your exported `CSSModulesLocalsConvention` type and the `applyLocalsConvention` unit tests were included as-is. Thanks again for the contribution!

PR #1040, opened 2026-07-29
fix(exports): scope inlined dependencies to configs with exports

### LINKED ISSUE #1039: inlinedDependencies leaks from configs that don't declare `exports`
> **Note:** this report was investigated and written up by an AI agent (Claude Opus 5, via Claude Code), reviewed by me before filing. The reproduction below was actually run against `tsdown@0.22.14`; the source references are from the installed dist.

Four files, no StackBlitz needed (happy to push a repo if you'd prefer one):

**`package.json`**

```json
{
  "name": "tsdown-inlined-repro",
  "version": "0.0.0",
  "private": true,
  "type": "module",
  "dependencies": { "pathe": "2.0.3" },
  "devDependencies": { "tsdown": "0.22.14" }
}
```

**`tsdown.config.ts`**

```ts
import { defineConfig } from "tsdown";

export default defineConfig([
  {
    // the library build: pathe stays external, nothing is inlined
    deps: { neverBundle: ["pathe"] },
    entry: ["src/index.ts"],
    exports: true,
    format: ["esm"],
  },
  {
    // a second, unrelated artifact: bundle everything, no `exports`
    deps: { alwaysBundle: [/.*/] },
    entry: ["src/standalone.ts"],
    format: ["cjs"],
    outDir: "dist/standalone",
  },
]);
```

**`src/index.ts`** — `export { join } from "pathe";`
**`src/standalone.ts`** — `import { join } from "pathe"; console.log(join("a", "b"));`

Then `pnpm install && pnpm tsdown`.

### What is expected?

`inlinedDependencies` should describe the artifacts the `exports` feature actually manages. The library build externalizes `pathe` (`dist/index.mjs` is literally `import { join } from "pathe"; export { join };`), so nothing is inlined and the field should be absent.

The second config's inlined deps leak into the field written on behalf of the first:

```json
"exports": {
  ".": "./dist/index.mjs",
  "./package.json": "./package.json"
},
"inlinedDependencies": {
  "pathe": "2.0.3"
}
```

`dist/index.mjs` still imports `pathe` at runtime, and `pathe` is still a declared `dependencies` entry — so the published metadata claims a package is physically vendored when a consumer will in fact resolve it normally.

### Any additional comments?

The merge in `bundleDone` is scoped per `package.json`, and unlike the chunk collection immediately below it, it doesn't filter to configs that declare `exports` (`src/features/pkg/…`, `dist/build-*.mjs`):

```js
const inlinedDeps = mergeInlinedDeps(ctx.bundles);   // every bundle for this pkg
for (const bundle of ctx.bundles) {
  if (!bundle.config.exports) continue;              // chunks: exports-configs only
  chunks[bundle.config.format] ||= [];
  chunks[bundle.config.format].push(...bundle.chunks);
}
await writeExports(exportsConfigs[0], chunks, inlinedDeps);
```

Moving `mergeInlinedDeps` to accumulate inside that loop (or filtering `ctx.bundles` by `config.exports` first) would scope it to the same bundles the exports map is generated from.

**Context for how I hit this:** the real case is a package that builds a normal ESM library plus a Node SEA binary from one config array. The SEA config is `deps: { alwaysBundle: [/.*/] }` and its output is excluded from `files`, so it never ships — but it contributed 34 entries to the published `package.json`, including six packages the library build explicitly lists in `deps.neverBundle`. Anything consuming this field for supply-chain or duplicate-dependency auditing would read that as vendored code.

There also doesn't appear to be an escape hatch short of dropping the field: `exports.inlinedDependencies: false` spreads `undefined` over the parsed package.json, so `JSON.stringify` drops the key entirely rather than leaving an existing value untouched. That's fine as a workaround here (the field is derived metadata, so there's nothing hand-authored to preserve) — noting it only because "disable" and "don't manage" aren't separable today.

lazerg:

### Description

When several build configs share one `package.json` and only some enable `exports`, the generated `inlinedDependencies` field pulled inlined deps from every config for that package, including ones that never contribute to the exports map. So a config that bundles everything (say a standalone binary excluded from `files`) leaked its inlined packages into the library's published `package.json`, even though the library externalizes those same packages.

The fix scopes the merge to the bundles the exports map is built from, so `inlinedDependencies` only lists deps that an exports config actually inlines.

### Linked Issues

Fixes #1039

### Additional context

Added a regression test with a two-config setup (one library config with `exports`, one bundle-everything config without) asserting the inlined dep from the second no longer appears in the package's `inlinedDependencies`.

### DIFF
--- a/tests/e2e.test.ts
+++ b/tests/e2e.test.ts
@@ -7,7 +7,6 @@ import {
   type InlineConfig,
   type UserConfig,
 } from '../src/config/index.ts'
-import { build } from '../src/index.ts'
 import { slash } from '../src/utils/general.ts'
 import { globalLogger } from '../src/utils/logger.ts'
 import { chdir, testBuild, writeFixtures } from './utils.ts'
@@ -541,16 +540,18 @@ describe('deps', () => {
     })
 
     test('should not include inlined deps from configs without exports', async (context) => {
-      const { testDir } = await writeFixtures(context, {
-        ...node_modules,
-        'index.ts': `export { lib } from 'my-lib'`,
-        'standalone.ts': `import { lib } from 'my-lib'\nconsole.log(lib)`,
-        'package.json': JSON.stringify({
-          name: 'test-pkg',
-          version: '1.0.0',
-          dependencies: { 'my-lib': '1.2.3' },
-        }),
-        'tsdown.config.ts': `export default [
+      const { testDir } = await testBuild({
+        context,
+        files: {
+          ...node_modules,
+          'index.ts': `export { lib } from 'my-lib'`,
+          'standalone.ts': `import { lib } from 'my-lib'\nconsole.log(lib)`,
+          'package.json': JSON.stringify({
+            name: 'test-pkg',
+            version: '1.0.0',
+            dependencies: { 'my-lib': '1.2.3' },
+          }),
+          'tsdown.config.ts': `export default [
           { entry: ['index.ts'], format: ['esm'], exports: true, dts: false },
           {
             entry: ['standalone.ts'],
@@ -560,14 +561,14 @@ describe('deps', () => {
             deps: { alwaysBundle: [/.*/] },
           },
         ]`,
+        },
+        options: {
+          entry: undefined,
+          config: 'tsdown.config.ts',
+          dts: false,
+        },
+        snapshot: false,
       })
-
-      const restoreCwd = chdir(testDir)
-      try {
-        await build({ config: testDir, logLevel: 'silent' })
-      } finally {
-        restoreCwd()
-      }
 
       const pkg = JSON.parse(
         await readFile(path.join(testDir, 'package.json'), 'utf8'),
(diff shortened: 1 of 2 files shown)


tests/e2e.test.ts
```diff
@@ -538,6 +539,41 @@ describe('deps', () => {
...
+            outDir: 'dist/standalone',
+            dts: false,
+            deps: { alwaysBundle: [/.*/] },
+          },
+        ]`,
+      })
+
+      const restoreCwd = chdir(testDir)
+      try {
+        await build({ config: testDir, logLevel: 'silent' })
```
sxzz: Use `testBuild` instead.

sxzz (changes requested):

tests/e2e.test.ts
lazerg: Done, switched it over to testBuild.

sxzz (approved):

PR #1041, opened 2026-07-29
fix: make `deps.dts` optional

ianzone:
This pull request makes a targeted update to the `ResolvedConfig` type in `src/config/types.ts` to improve type flexibility for the `deps` property. The change omits the `dts` property from `ResolvedDepsConfig` and reintroduces it as an optional property, allowing for more nuanced configuration.

- **Type flexibility improvement:**
  * Updated the `deps` property in the `ResolvedConfig` type to omit the `dts` property from `ResolvedDepsConfig` and redefine it as optional, enhancing configurability for dependency settings.


sxzz:
Seems like written by AI. Please respect PR template.

PR #1058, opened 2026-08-21
fix(cli): apply `--deps.never-bundle` by camel-casing dotted flag keys

dylanpulver:
### AI usage

- [x] AI was used: Claude Code + Opus 5
  - [ ] I have carefully reviewed the AI-generated content myself.

### Description

`--deps.never-bundle` has been a no-op since the `deps` namespace rename (7f50966). cac's `camelcaseOptionName` camel-cases only the first segment of a dotted option name, so the flag parses to `deps['never-bundle']` while `resolveDepsConfig` reads `deps.neverBundle`. Nothing errors, because the command sets `allowUnknownOptions`.

On `main`, with `leftpad` present in `node_modules`:

```
tsdown --deps.never-bundle leftpad   # leftpad inlined into the bundle
tsdown --deps.neverBundle leftpad    # leftpad external
tsdown --external leftpad            # leftpad external (deprecated flag)
```

Renaming the option to `--deps.neverBundle` was the alternative. That leaves the kebab spelling silently doing nothing, and the kebab spelling is what `docs/reference/cli.md` and `skills/tsdown/references/` document, so this normalizes the parsed keys and both spellings work.

Normalization is scoped to `flags.deps` at one level. `--env.*` is excluded since its keys are user-supplied variable names. A future dotted option in another namespace would need the same handling.

`src/cli.test.ts` parses through the real `cli` instance, which required exporting it. The first case fails on `main`; the other two pin current behavior.

### DIFF
--- a/src/cli.test.ts
+++ b/src/cli.test.ts
@@ -1,0 +1,27 @@
+import { expect, test } from 'vitest'
+import { cli, normalizeFlags } from './cli.ts'
+import type { UserConfig } from './config.ts'
+
+function parseFlags(...argv: string[]): UserConfig {
+  cli.parse(['node', 'tsdown', ...argv], { run: false })
+  return normalizeFlags(cli.options as UserConfig)
+}
+
+test('--deps.never-bundle maps to deps.neverBundle', () => {
+  expect(parseFlags('--deps.never-bundle', 'lodash').deps).toEqual({
+    neverBundle: 'lodash',
+  })
+})
+
+test('--deps.neverBundle keeps working', () => {
+  expect(parseFlags('--deps.neverBundle', 'lodash').deps).toEqual({
+    neverBundle: 'lodash',
+  })
+})
+
+test('--env.* keys are left verbatim', () => {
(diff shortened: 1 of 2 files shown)


sxzz:
Thanks for tracking this down, the diagnosis is right.

I went with a broader fix though. `camelcaseOptionName` applies to every parsed key, so all nested kebab flags are ignored, not just this one: `--deps.always-bundle`, `--deps.dts.never-bundle`, `--dts.emit-dts-only`, `--exports.dev-exports`. The docs also say kebab and camelCase are equivalent for any option, listed or not. Scoping it to `flags.deps` at one level leaves the rest broken, and every future nested option would need another special case.

`env` is also not the only namespace with user-defined keys. `define`, `alias`, `entry` and `loader` are Records too, so `--alias.my-lib ./src/my-lib.ts` would break once those keys get camelized.

One more thing: exporting `cli` plus a function marked `@private` to make it testable isn't something I want in the codebase. b6884a4 normalizes right after `cli.parse()` inside `parseCLI()`, which is a regular internal function `runCLI()` calls, so the test goes through the same path production does.

Your test cases live on in `src/cli.test.ts`. Closing as superseded, thanks for the report.

Issue #1067, opened 2026-09-10
workspace: include globs match directories without package.json, and the resulting "Cannot find entry" error names the root config instead of the offending directory


relanse:
### Reproduction link or steps

Repro repo: https://github.com/relanse/tsdown-ghost-dir-repro

```sh
git clone https://github.com/relanse/tsdown-ghost-dir-repro
cd tsdown-ghost-dir-repro
npx tsdown@latest
```

Layout:

- `tsdown.config.ts` — `workspace: ['pkgs/*']` with a root-level `entry: ['src/{index,ghost}.ts']`
- `pkgs/real/` — a real package whose `src/index.ts` matches the entry glob
- `pkgs/ghost/` — a directory with **no `package.json` and no tsdown config** (only a `.gitkeep`; e.g. a stale leftover after a package rename)

### What is expected?

Either:

- Directories that contain neither a `package.json` nor a tsdown config file are skipped, and `pkgs/real` builds successfully; or
- The error names the offending directory (`pkgs/ghost`) instead of the root config.

### What is actually happening?

The whole build fails with an error labeled with the **root** config's name, and the message contains no path:

```
ERROR  Error: [repro-root] Cannot find entry: ["src/{index,ghost}.ts"]
```

`pkgs/real/src/index.ts` exists and matches the entry glob, but the build never reaches it: the package-less `pkgs/ghost` also gets built — it has no config of its own, so it inherits the root `entry` via `mergeConfig(normalized, config)` — and entry resolution matches nothing inside that directory.

In `src/config/workspace.ts`, `include: 'auto'` globs `**/package.json` so it only ever finds real packages, but a user-provided `include` array globs directories (`onlyDirectories: true`) with no package-ness check. The error's `nameLabel` (`src/features/entry.ts`) then carries the merged config's inherited root name.

### Any additional comments?

Verified with tsdown 0.23.0 and 0.22.2, Node v24.16.0, Windows 10.

We hit this in the deepseek-harness monorepo (~270 packages), whose root tsdown config drives workspace builds this way (`workspace: ['vendor/*', 'packages/*/*', 'apps/cli']`). One stale directory left behind after a package rename — Windows file locks commonly prevent git from fully removing directories — broke `pnpm run build` for the whole monorepo with an error pointing at the root package. Locating the actual culprit took a long time precisely because the message carries no path.

Suggested fix (either or both):

[...]

PR #1069, opened 2026-09-11
fix(workspace): skip include-matched directories without a package.json

JiangYueYun:
### AI usage

- [x] AI was used: DeepSeek Harness (deepseek-v4-pro)
  - [x] I have carefully reviewed the AI-generated content myself.

### Description

`workspace.include` is a list of glob patterns, expanded with `onlyDirectories: true`, so a directory
that is not a package — a directory left behind by a package rename, for example — becomes a
workspace entry. It has no config of its own, so it inherits the root config's `entry` and `name`,
and the build then fails with `[root-name] Cannot find entry: [...]`: the message names the root
package and carries no path.

This brings the `include` branch in line with `auto`, which discovers `**/package.json` and so never
matches a non-package: directories matched by the globs that own no `package.json` are skipped, and
reported instead of being passed over quietly.

    WARN  workspace include matched 2 directories without a package.json; skipping: packages/a, packages/b

### Source report

Reproduced from a real monorepo: deepseek-ai/deepseek-harness discussion #6089
(https://github.com/deepseek-ai/deepseek-harness/discussions/6089). `pnpm run build:lib:host` fails
with `[@deepseek-ai/dsh-root] Cannot find entry: ["lib/types/{index,invariant,startup}.js"]` on a tree
that holds a directory without a `package.json` under `packages/*/*`, while a clean checkout of the
same revision builds 249 packages without a root `lib/types`.

### Linked Issues

Refs #1067. #1068 covers the other half — naming the offending directory in the error message; this
covers "a directory that is not a package must not join the workspace".

### Tests

`tests/e2e.test.ts` gains a workspace whose include glob matches one real package and one directory
holding only `.gitkeep`: it fails on `main` (`Build failed with 1 error`) and passes with the change,
together with its snapshot. `eslint --max-warnings 0` and `tsc --noEmit` are clean on the touched
files.


JiangYueYun:
Closing this: #1067 already carries the diagnosis, which I hit independently while building a monorepo, and #1068 covers the error-message side. The implementation here is yours to take, rewrite, or drop — I won't be iterating on the patch.

Issue #327, opened 2025-06-17
`Unbundle` mode combined with `rolldown.preserveModulesRoot`


jevin98:
### Clear and concise description of the problem

I see that `rolldown` already supports `preserveModulesRoot` configuration. Should `tsdown` also add a configuration?

Or is it set directly through `outputOptions`?

### Suggested solution

-

### Validations

- [x] Follow our [Code of Conduct](https://github.com/sxzz/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/sxzz/contribute).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sxzz:
tsdown will automatically set `preserveModulesRoot`. If you wish to customize this setting, use `outputOptions`.

PR #1072, opened 2026-09-13
fix(entry): honor custom root for object entry

giaBaoJS:
### AI usage

- [x] AI was used: Claude Code + Claude Opus 5
  - [ ] I have carefully reviewed the AI-generated content myself.

### Description

`root` is documented as "Specifies the root directory of input files, similar to TypeScript's `rootDir`. This determines the output directory structure." with no restriction on the entry form, and the array and glob entry paths honor it. Object entries silently ignore it.

`toObjectEntry` takes `root` and forwards it to `resolveArrayEntry`, but calls `resolveObjectEntry(entry, cwd)` without it. `resolveObjectEntry` has no `root` parameter, so it always returns `lowestCommonAncestor(...)`. That value ends up in `config.root`, which feeds `preserveModulesRoot: unbundle ? config.root : undefined`, so the output layout in unbundle mode ignores the option.

Reproduction, same `root` and same files, only the entry form differs:

```js
// tsdown.config.ts
export default {
  entry: { index: 'src/index.ts' },   // vs. ['src/index.ts']
  root: '.',
  unbundle: true,
}
```

with `src/index.ts` importing `src/deep/nested/shared.ts`:

```
array entry  -> dist/src/index.mjs, dist/src/deep/nested/shared.mjs   (root honored)
object entry -> dist/index.mjs,     dist/deep/nested/shared.mjs       (root ignored)
```

The fix threads `root` into `resolveObjectEntry` and keeps the computed ancestor as the fallback, mirroring `resolveArrayEntry`. Entry keys are untouched: for object entries they are user-provided, so unlike the array path the root must not rename them.

Three tests next to the existing `array entry with custom root` / `glob entry with custom root` cases: a plain object entry with a custom root, an object entry with a glob key and a custom root, and an object entry without a root to pin the existing lowest-common-ancestor fallback. The first two fail on `main` and pass here; the third passes on both and guards against the fallback being dropped.

Related: #327 was closed as `not_planned` with "tsdown will automatically set `preserveModulesRoot`. If you wish to customize this setting, use `outputOptions`". That predates the `root` option. This PR does not add a new knob, it makes the existing documented one work for a second entry form.

### Linked Issues

None.

### Additional context

Local run: `pnpm run build` then `pnpm run test` gives 530 passed / 4 expected fail / 4 skipped (527 passed before this change, plus the 3 new tests), `pnpm run typecheck` and `pnpm run lint` clean.


sxzz (approved):

sxzz:
Merged, thanks!

PR #1074, opened 2026-09-14
feat: add JSON schema for config

### LINKED ISSUE #679: JSON Schema for tsdown.config.json

Tsdown already supports configuring with a JSON file, but currently does not provide a JSON Schema. Adding a JSON Schema would improve the developer experience by enabling better editor auto-completion and validation when using JSON to configure Tsdown.

### Suggested solution

Provide a JSON Schema for tsdown, just like [swc](https://swc.rs/schema.json), to enhance configuration editor experience.

### Alternative

_No response_

### Additional context

_No response_

### Validations

Sandbye:
### AI usage

- [x] AI was used: Codex (gpt-5.6-luna High) 
  - [x] I have carefully reviewed the AI-generated content myself.

### Description

Attempt at adding a generated JSON Schema for `tsdown.config.json`.

**Generation**
The schema was generated from tsdown’s `UserConfig` TypeScript type using the TypeScript compiler API.

It preserves:

  - Property descriptions from JSDoc
  - `@default` values when they are JSON-compatible literals
  - `@deprecated` annotations
  - Reusable `$defs` and `$ref` definitions

Function-only options, plugins, and callbacks are omitted because they cannot be represented in JSON.

The generator is currently kept separate from this PR. If preferred, it can be added to the repository later to regenerate the schema automatically.

### Linked Issues
fix #679

### Additional context

Verified with:
  - pnpm lint
  - pnpm typecheck
  - pnpm test run


sxzz:
The JSON Schema is too large to maintain manually, so we need automated tooling to keep it up to date. Please include the generator in this PR; without it, ongoing maintenance would not be feasible.

The schema file is also quite large and would increase the npm package size. I'd prefer to keep it in the GitHub repository and let users load it via a URL, rather than include it in the published npm package.

Sandbye:
Yeah, i totally agree. Will follow up on that.

sxzz:
Actually, I'd prefer to have the generator produce a Zod (or Zod-like) schema first, then convert it to JSON Schema. This would make the code more intuitive to maintain and the process more reliable.

sxzz:
Since Rolldown uses Valibot, I'd actually prefer Valibot over Zod here.

Sandbye:
Agreed. I’ll rework the generator to produce Valibot schemas first and convert those to JSON Schema, then include the generator and generated schema in the PR. I’ll also keep the generated schema out of the npm package and serve it from the repository instead. :)

scripts/generate-schema.ts
```diff
@@ -0,0 +1,49 @@
...
+  type SchemaRoot,
+} from './schema-generator/index.ts'
+
+const args = process.argv.slice(2)
+const cwd = getArgument('--cwd') ?? process.cwd()
+const sourcePath = getArgument('--source') ?? 'src/config/types.ts'
+const typeName = getArgument('--type') ?? 'UserConfig'
+const outputPath = getArgument('--out') ?? path.resolve(cwd, 'schema.json')
+const tsconfigPath = getArgument('--tsconfig')
+const rootValue = getArgument('--root') ?? 'object-or-array'
```
sxzz: Use `cac` instead.

scripts/schema-generator/index.ts
```diff
@@ -0,0 +1,538 @@
...
+    definitions: generated.definitions,
+    errorMode: 'throw',
+  })
+}
+
+export function writeSchema(options: WriteSchemaOptions): void {
+  const { outputPath, ...generateOptions } = options
+  writeFileSync(
+    path.resolve(outputPath),
+    `${JSON.stringify(generateSchema(generateOptions), null, 2)}\n`,
```
sxzz: Use prettier to format

package.json
```diff
@@ -80,6 +79,8 @@
     "test": "vitest",
     "typecheck": "tsc --noEmit",
     "format": "prettier --cache --write .",
+    "schema:generate": "node ./scripts/generate-schema.ts --out ./schema.json && prettier --write ./schema.json",
+    "schema:check": "node ./scripts/generate-schema.ts --check --out ./schema.json",
```
sxzz:
```json
    "schema:generate": "node ./scripts/generate-schema.ts",
    "schema:check": "node ./scripts/generate-schema.ts -c",
```

- Add Prettier integration in the script
- Add `-c` shortcut for `--check`.
- Remove default `out` flag

tsconfig.json
```diff
@@ -31,6 +31,7 @@
   },
   "include": [
     "src",
+    "scripts/schema-generator",
```
sxzz: Prefer `"scripts"`

sxzz:
Also run `schema:generate` script in `autofix.yml`.

Sandbye:
**Question**
I found that function/API-only types are not limited to the root-level hooks and customLogger options. Nested options such as format-specific hooks/loggers, minify.mangleProps, and visitor APIs also become empty schemas `({})`.

Should the JSON-specific schema omit all function/API-only options recursively, or should these remain unconstrained?

sxzz:
I'd prefer to keep these options unconstrained in the JSON Schema, including nested ones, rather than omit them.

Sandbye:
> I'd prefer to keep these options unconstrained in the JSON Schema, including nested ones, rather than omit them.

Cool, then this PR is ready for final review. Nothing else from me.

silashoeyer:
looks good to me

sxzz:
LGTM. Maybe we can also introduce a new package `@tsdown/schema`, providing valibot & JSON schema to AI agents.

Sandbye:
> LGTM. Maybe we can also introduce a new package `@tsdown/schema`, providing valibot & JSON schema to AI agents.

Great! Should @tsdown/schema be a follow-up issue, or would you rather have it included in this pr?

sxzz:
I'm okay for both. It's up to you. BTW, I'm currently on vacation, and this PR will be delayed by few days.

Sandbye:
Have a nice vacation 🌴
I'll look into it when i have time next week.

Sandbye:
@sxzz  
I’ll leave @tsdown/schema as a follow-up so this PR stays focused on adding the JSON Schema and generator. Thanks for the suggestion!

